PR #11, opened 2022-09-20
Address some comments and add more tests

nothing0012:
- Add frozenBaseURI event and freezeBaseURI so that we can freeze the metadata
- We cannot increase the max supply once it's initialized from the constructor
- Refactored the naming for _stageMintedCountsPerWallet and _stageMintedCounts
- Increased the coverage

image

There's still a TODO: fix the 'Provider not found' error

### DIFF
--- a/test/erc721m.test.ts
+++ b/test/erc721m.test.ts
@@ -1,3 +1,4 @@
+import { SignerWithAddress } from '@nomiclabs/hardhat-ethers/signers';
 import chai, { assert, expect } from 'chai';
 import chaiAsPromised from 'chai-as-promised';
 import { ethers } from 'hardhat';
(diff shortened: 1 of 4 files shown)


test/erc721m.test.ts
```diff
@@ -574,6 +668,34 @@ describe('ERC721M', function () {
...
+        value: ethers.utils.parseEther('2.5'),
+      });
+
+      expect(await contract.tokenURI(0)).to.equal('base_uri_0');
+      expect(await contract.tokenURI(1)).to.equal('base_uri_1');
+
+      await contract.setBaseURI('base_uri_again_');
+      expect(await contract.tokenURI(0)).to.equal('base_uri_again_0');
+
+      // readonlyContract should not be able to set active stage
```
mi-yu, line 693: nit: comment should read `not be able to set baseURI`

mi-yu (approved):

PR #17, opened 2022-09-26
Add cosign server

nothing0012:
Add an open source implementation of the cloudflare workers to support cosign.


cosign-server/src/index.ts
```diff
@@ -0,0 +1,157 @@
+import { ethers } from "ethers";
```
mi-yu, line 1: we can use https://www.npmjs.com/package/@ethersproject/solidity to reduce bundle size

PR #20, opened 2022-09-26
Support crossmint

channing-magiceden:
- Add `crossmint` function
- `crossmintAddress` needs to be set in order to enable it
- Refactor mint functions so they share the validation logic

File           |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
---------------|----------|----------|----------|----------|----------------|
 contracts/    |      100 |     91.3 |      100 |      100 |                |
  ERC721M.sol  |      100 |     91.3 |      100 |      100 |                |
  IERC721M.sol |      100 |      100 |      100 |      100 |                |
All files      |      100 |     91.3 |      100 |      100 |                |

https://www.loom.com/share/9d646fe717aa462497dd801bd4504d19

### DIFF
--- a/contracts/ERC721M.sol
+++ b/contracts/ERC721M.sol
@@ -12,6 +12,9 @@ import "./IERC721M.sol";
 contract ERC721M is IERC721M, ERC721AQueryable, Ownable {
     using ECDSA for bytes32;
 
+    address private constant CROSSMINT_ADDRESS =
+        0xdAb1a1854214684acE522439684a145E62505233;
+
     bool private _mintable;
     string private _currentBaseURI;
     uint256 private _activeStage;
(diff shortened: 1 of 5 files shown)


test/erc721m.test.ts
```diff
@@ -664,6 +664,34 @@ describe('ERC721M', function () {
...
+        [7],
+      );
+      await contract.setMintable(true);
+
+      const crossMint = contract.crossMint(7, recipient.address, proof, 0, '0x00', {
+        value: ethers.utils.parseEther('3.5'),
+      });
+
+      await expect(crossMint).to.be.revertedWith('CrossmintOnly');
+    });
```
nothing0012, line 693: can we test the happy code path? I think hardhat might have something to stub a caller private key to be the crossmint address.

contracts/ERC721M.sol
```diff
@@ -12,6 +12,9 @@ import "./IERC721M.sol";
 contract ERC721M is IERC721M, ERC721AQueryable, Ownable {
     using ECDSA for bytes32;
 
+    address private constant CROSSMINT_ADDRESS =
```
mi-yu, line 15: Just thinking, in case of some emergency like crossmint private key is compromised, do we want to be able to update this?

nothing0012: yeah, I think make it configurable also helps with testing.

channing-magiceden: Done. Made it configurable like cosigner.

test/erc721m.test.ts
channing-magiceden: Done.

contracts/ERC721M.sol
```diff
@@ -197,7 +207,32 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable {
...
+        _mintInternal(qty, to, proof, timestamp, signature);
+    }
+
+    function _mintInternal(
+        uint32 qty,
+        address to,
+        bytes32[] calldata proof,
+        uint256 timestamp,
+        bytes calldata signature
+    ) internal canMint hasSupply(qty) {
```
nothing0012, line 235: assertValidCosign(msg.sender, qty, timestamp, signature); for this line, I'm thinking do we need to change this to `to`

nothing0012: yeah, second thought is that yeah, it should be msg.sender, that we cosign to stamp for the msg sender, not on behalf of the `to`, even though for the `mint()` that msg.sender == to.

nothing0012: this means that if we turn on cosign, crossmint also need to talk to our api to get the signature first. @channing-magiceden let's chat with crossmint to see if they can support it, otherwise we might need to make this cosign check optional

nothing0012:
@channing-magiceden  going to merge this so that we can kickoff halborn's audit

PR #22, opened 2022-09-27
Add BucketAuction contract

nothing0012:
TODO: add more test coverage, try to hit 100% lines


contracts/BucketAuction.sol
```diff
@@ -0,0 +1,282 @@
...
+    /**
+     * @notice place a bid in ETH or add to your existing bid. Calling this
+     *   multiple times will increase your bid amount. All bids placed are final
+     *   and cannot be reversed.
+     */
+    function bid() external payable isAuctionActive nonReentrant cannotMint {
+        User storage bidder = _userData[msg.sender]; // get user's current bid total
+        uint256 contribution_ = bidder.contribution; // bidder.contribution is uint216
+        unchecked {
+            // does not overflow
```
channing-magiceden, line 86: Why do we use uint216 for contribution and why it doesn't overflow?

contracts/BucketAuction.sol
```diff
@@ -0,0 +1,282 @@
...
+        bidder.contribution = uint216(contribution_);
+        emit Bid(msg.sender, msg.value, contribution_, address(this).balance);
+    }
+
+    /**
+     * @notice set the minimum contribution required to place a bid
+     * @dev set this price in wei, not eth!
+     * @param minimumContributionInWei new price, set in wei
+     */
+    function setMinimumContribution(uint256 minimumContributionInWei)
```
channing-magiceden, line 100: Does it make sense to limit this call before the auction is active?

contracts/BucketAuction.sol
```diff
@@ -0,0 +1,282 @@
...
+        _internalMint(to, n);
+    }
+
+    /**
+     * @notice get the number of tokens purchased by an address, after the
+     *   clearing price has been set.
+     * @param a address to query.
+     */
+    function amountPurchased(address a) public view returns (uint256) {
+        return _userData[a].contribution / _price;
```
channing-magiceden, line 160: qq: the # tokens are decided by `contribution` and `_price` here. But do we need to support the case that user explicitly bid for X tokens with Y contribution?

e.g. I bid 1 token for 200 wei. The price is set to 100. I just get 1 token and 100 wei refund.

contracts/BucketAuction.sol
```diff
@@ -0,0 +1,282 @@
...
+        _minimumContributionInWei = minimumContributionInWei;
+        emit SetMinimumContribution(minimumContributionInWei);
+    }
+
+    /**
+     * @notice set the clearing price after all bids have been placed.
+     * @dev set this price in wei, not eth!
+     * @param priceInWei new price, set in wei
+     */
+    function setPrice(uint256 priceInWei)
```
channing-magiceden, line 113: Should we only allow setting price once? 

Just read through the contract. I think the one vulnerability is allowing owner changing price. I am not sure if this is ok but just want to raise this.

e.g. owner can reset price so he can send more tokens to specific bidders.

channing-magiceden, line 113: Another question: how's the price determined? If based on total contribution, do we need a method to return sum of contribution?

channing-magiceden: All refund / token calculation are based on current price. Some users may lose fund if their `refundClaimed` is set with different price.

contracts/BucketAuction.sol
nothing0012: no, that's not bucket auction - that's just auction with multiple bids

contracts/BucketAuction.sol
nothing0012: maybe we make it impossible to change price once there's refund happened, good catch!

nothing0012: we need to get this info (maybe https://thegraph.com/en/) can be helpful here, and calculate the optimal price offchain. can you work on this part after this PR?

contracts/BucketAuction.sol
nothing0012: compact the User struct storage

channing-magiceden: TODO: let's test if using 216 helps

contracts/BucketAuction.sol
channing-magiceden: Resolved.

channing-magiceden (approved):

PR #23, opened 2022-09-28
Use changeEtherBalances to test withdraw

macelai:
The problem was that it was using `changeEtherBalance` that expects only an account. Now it uses `changeEtherBalances` that expects an method call.


nothing0012 (approved):
awesome!

PR #26, opened 2022-10-02
Use cosign + stage timestamps to automatically determine active stage

mi-yu:
## Breaking changes
- stages now have `startTimeUnixSeconds` and `endTimeUnixSeconds`
- if cosign is enabled, timestamp is used to determine the active stage
  - timestamp must be within the last 60 seconds according to `block.timestamp`, so you cannot use an old signature from cosign-server to mint a previous stage after its ended
  - stages have at least 60 seconds between them

Working on 100% branch coverage
Screen Shot 2022-10-01 at 10 28 18 PM

### DIFF
--- a/contracts/ERC721M.sol
+++ b/contracts/ERC721M.sol
@@ -6,7 +6,6 @@ import "@openzeppelin/contracts/utils/cryptography/MerkleProof.sol";
 import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
 import "@openzeppelin/contracts/utils/cryptography/SignatureChecker.sol";
 import "erc721a/contracts/extensions/ERC721AQueryable.sol";
-import "hardhat/console.sol";
 import "./IERC721M.sol";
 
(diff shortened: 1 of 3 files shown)


madergaser (approved):

contracts/ERC721M.sol
```diff
@@ -371,4 +404,32 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable {
...
     }
+
+    function getActiveStageFromTimestamp(uint256 timestamp)
+        public
+        view
+        returns (uint256)
+    {
+        for (uint256 i = 0; i < _mintStages.length; i++) {
+            if (
+                timestamp >= _mintStages[i].startTimeUnixSeconds &&
```
madergaser, line 415: nit: since we check on write path, read path only needs to check `<`

mi-yu: i actually think we need the `>=` check in case we are in between stages, or stage hasn't started yet

PR #28, opened 2022-10-03
Initial commit for dutch auction

nothing0012:
- Add Dutch Auction

File                 |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
---------------------|----------|----------|----------|----------|----------------|
 contracts/          |      100 |    90.57 |      100 |      100 |                |
  BucketAuction.sol  |      100 |    91.03 |      100 |      100 |                |
  DutchAuction.sol   |      100 |     87.5 |      100 |      100 |                |
  ERC721M.sol        |      100 |    91.49 |      100 |      100 |                |
  IBucketAuction.sol |      100 |      100 |      100 |      100 |                |
  IDutchAuction.sol  |      100 |      100 |      100 |      100 |                |
  IERC721M.sol       |      100 |      100 |      100 |      100 |                |
---------------------|----------|----------|----------|----------|----------------|
All files            |      100 |    90.57 |      100 |      100 |                |


contracts/DutchAuction.sol
```diff
@@ -0,0 +1,176 @@
...
+        uint256 startAmountInWei,
+        uint256 endAmountInWei,
+        uint256 startTime,
+        uint256 endTime,
+        bool roundUp
+    ) external onlyOwner {
+        if (startTime == 0 || startTime >= endTime)
+            revert InvalidStartEndTime();
+        if (startAmountInWei == 0) revert InvalidAmountInWei();
+
```
nothing0012, line 61: we might need to add some guard rails for owner, that if the DA already started, might want to prevent some config fields from being changed.

contracts/DutchAuction.sol
```diff
@@ -0,0 +1,176 @@
...
+        Config memory config;
+        config = Config({
```
nothing0012, line 63: might need to just merge these lines.

contracts/DutchAuction.sol
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    function getSettledPriceInWei() external view returns (uint256) {
+        return _settledPriceInWei;
+    }
+
+    function getCurrentPriceInWei() public view returns (uint256) {
+        Config memory config = _config; // storage to memory
+        uint256 amount;
+        bool roundUp = config.roundUp;
```
nothing0012, line 84: might want to move the declaration of these two into the if block

contracts/DutchAuction.sol
```diff
@@ -0,0 +1,176 @@
...
+import "@openzeppelin/contracts/token/common/ERC2981.sol";
+import "erc721a/contracts/ERC721A.sol";
+
+import "./IDutchAuction.sol";
+import "./ERC721M.sol";
+
+contract DutchAuction is IDutchAuction, ERC721M, ReentrancyGuard {
+    bool private immutable _refundable;
+    uint256 private _settledPriceInWei;
+    Config private _config;
```
mi-yu, line 17: nit: rename to `DucthAuctionConfig` or similar

contracts/DutchAuction.sol
```diff
@@ -0,0 +1,176 @@
...
+
+    function setConfig(
+        uint256 startAmountInWei,
+        uint256 endAmountInWei,
+        uint256 startTime,
+        uint256 endTime,
+        bool roundUp
+    ) external onlyOwner {
+        if (startTime == 0 || startTime >= endTime)
+            revert InvalidStartEndTime();
```
mi-yu, line 59: nit: can include the `startTime` and `endTime` as parameters in the error https://blog.soliditylang.org/2021/04/21/custom-errors/

contracts/DutchAuction.sol
```diff
@@ -0,0 +1,176 @@
...
+
+    function claimRefund() external nonReentrant {
+        Config memory config = _config;
+        if (!_refundable) revert NotRefundable();
+        if (config.endTime > block.timestamp) revert NotEnded();
+
+        User storage user = _userData[msg.sender];
+        if (user.refundClaimed) revert UserAlreadyClaimed();
+        user.refundClaimed = true;
+        uint256 refundInWei = user.contribution -
```
mi-yu, line 167: nit: parens here for clarity?

contracts/IDutchAuction.sol
```diff
@@ -0,0 +1,28 @@
...
+        uint32 tokensBidded; // cumulative sum of bidded tokens
+        bool refundClaimed; // has user been refunded yet
+    }
+
+    struct Config {
+        uint256 startAmountInWei;
+        uint256 endAmountInWei;
+        uint256 startTime;
+        uint256 endTime;
+        bool roundUp;
```
mi-yu, line 24: do we need this to be an option? maybe we can just always round up?

mi-yu (approved):
overall lgtm, pending tests

contracts/DutchAuction.sol
```diff
@@ -0,0 +1,176 @@
...
+        emit Bid(msg.sender, qty, price);
+    }
+
+    function claimRefund() external nonReentrant {
+        Config memory config = _config;
+        if (!_refundable) revert NotRefundable();
+        if (config.endTime > block.timestamp) revert NotEnded();
+
+        User storage user = _userData[msg.sender];
+        if (user.refundClaimed) revert UserAlreadyClaimed();
```
channing-magiceden, line 165: What if user

bid -> claimRefund -> bid -> claimRefund?

The second claimRefund would fail

channing-magiceden: Proposed solution is to flip `user.refundClaimed` to true in `bid()`.

contracts/DutchAuction.sol
```diff
@@ -0,0 +1,176 @@
...
+            // Skip underflow checks as startTime <= block.timestamp < endTime.
+            unchecked {
+                // Derive the duration for the order and place it on the stack.
+                duration = config.endTime - config.startTime;
+
+                // Derive time elapsed since the order started & place on stack.
+                elapsed = block.timestamp - config.startTime;
+
+                // Derive time remaining until order expires and place on stack.
+                remaining = duration - elapsed;
```
channing-magiceden, line 101: Remaining will be underflow because `elapsed` could be > `duration`.

We should either:
1. check `validTime` to call this fn
2. return `startAmountInWei` if before start time, and return `endAmountInWei` if after end time

wdyt?

mi-yu: prefer option 2, so the UI can correctly display time even after the sale has ended

contracts/DutchAuction.sol
mi-yu: user shouldn't be able to bid again after claiming refund, since you can only bid before `endTime` and you can only `claimRefund` after `endTime`

channing-magiceden: Oh right, this is a false alarm.

contracts/DutchAuction.sol
channing-magiceden: +1. Updated.

contracts/DutchAuction.sol
nothing0012: yeah, the entTime gates that

nothing0012: I'm actually thinking if we just adopt this startTime/endTime for bucket auction as well, so it's consistent and we don't need to set additional 'isActive' stage, so to minimize the operation overhead

contracts/IDutchAuction.sol
nothing0012: yes, good idea

contracts/DutchAuction.sol
nothing0012: I think it's fine, since the Config is specific to the IDutchAuction interface

PR #29, opened 2022-10-04
Pack stage struct to save storage/gas

mi-yu:
- Saves ~50% gas on `setStages` (~490k -> 250k)
- New limits:
  - max price = 1.2M ETH
  - max stage supply = 16.7 million

### DIFF
--- a/contracts/IERC721M.sol
+++ b/contracts/IERC721M.sol
@@ -25,22 +25,22 @@ interface IERC721M {
     error WithdrawFailed();
 
     struct MintStageInfo {
-        uint256 price;
+        uint80 price;
         uint32 walletLimit; // 0 for unlimited
         bytes32 merkleRoot; // 0x0 for no presale enforced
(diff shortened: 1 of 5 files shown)


contracts/IERC721M.sol
```diff
@@ -25,22 +25,22 @@ interface IERC721M {
...
-        uint256 price;
+        uint80 price;
         uint32 walletLimit; // 0 for unlimited
         bytes32 merkleRoot; // 0x0 for no presale enforced
-        uint256 maxStageSupply; // 0 for unlimited
-        uint256 startTimeUnixSeconds;
-        uint256 endTimeUnixSeconds;
+        uint24 maxStageSupply; // 0 for unlimited
+        uint64 startTimeUnixSeconds;
+        uint64 endTimeUnixSeconds;
```
nothing0012, line 33: looking good, I think the writer is from the owner, for double checking the boundary of these fields

nothing0012 (approved):

PR #33, opened 2022-10-05
Add view methods to IERC721M

mi-yu:
- We can implement read proxies for arbitrary minting contracts with some pattern like this:

```
contract SomeERC721 {}

contract SomeERC721Adapter is IERC721M {}
```

Also fix timestamp sizes in test

### DIFF
--- a/contracts/IERC721M.sol
+++ b/contracts/IERC721M.sol
@@ -90,7 +90,7 @@ interface IERC721M is IERC721AQueryable {
     function assertValidCosign(
         address minter,
         uint32 qty,
-        uint256 timestamp,
+        uint64 timestamp,
         bytes memory signature
     ) external view;
 }
--- a/contracts/IERC721M.sol
+++ b/contracts/IERC721M.sol
@@ -1,7 +1,9 @@
 //SPDX-License-Identifier: MIT
 pragma solidity ^0.8.4;
 
-interface IERC721M {
+import "erc721a/contracts/extensions/IERC721AQueryable.sol";
+
+interface IERC721M is IERC721AQueryable {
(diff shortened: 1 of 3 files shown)


contracts/IERC721M.sol
```diff
@@ -52,4 +54,43 @@ interface IERC721M {
...
+
+    function getActiveStageFromTimestamp(uint64 timestamp)
+        external
+        view
+        returns (uint256);
+
+    function assertValidCosign(
+        address minter,
+        uint32 qty,
+        uint256 timestamp,
```
channing-magiceden, line 93: Make all timestamp `uint64`?

contracts/IERC721M.sol
```diff
@@ -52,4 +54,43 @@ interface IERC721M {
...
+
+    function getMaxMintableSupply() external view returns (uint256);
+
+    function getMintable() external view returns (bool);
+
+    function totalMintedByAddress(address a) external view returns (uint256);
+
+    function getTokenURISuffix() external view returns (string memory);
+
+    function getStageInfo(uint256 index)
```
channing-magiceden, line 74: Do we need uint256 for stage index?

contracts/IERC721M.sol
mi-yu: yes let me consolidate everything

contracts/IERC721M.sol
mi-yu: I believe one storage slot is 32 bytes (256 bits), so anything smaller will still require 1 slot allocated https://docs.soliditylang.org/en/v0.8.15/internals/layout_in_storage.html

channing-magiceden (approved):

PR #36, opened 2022-10-07
WIP fix cosign

nothing0012:


contracts/ERC721M.sol
```diff
@@ -31,6 +33,8 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
         private _stageMintedCountsPerWallet;
     mapping(uint256 => uint256) private _stageMintedCounts;
 
+    mapping(address => Counters.Counter) private _cosignNonces;
```
nothing0012, line 36: just use the minted count as the nonce, so we can save the storage of this mapping here

contracts/ERC721M.sol
```diff
@@ -397,7 +408,9 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
                     minter,
                     qty,
                     _cosigner,
-                    timestamp
+                    timestamp,
+                    _chainID(),
+                    getCosignNonce(minter)
```
nothing0012, line 413: double check if burn token will decrease the mintedCount mapping from erc721a

nothing0012: EIP712 supports https://eips.ethereum.org/EIPS/eip-712#definition-of-domainseparator

which means that if a contract has more than 1 domain that needs signature, we might want to add the domain into the hashing seeds.

but since this signature is only used for minting, the default domain is 'minting'

PR #41, opened 2022-10-21
Pass minted tokenIds as last param of callback function as temp solution

channing-magiceden:
Pass minted tokenIds as last param of callback function as temp solution


mi-yu (approved):

mi-yu:
closing this one as we need to re-evaluate the callback feature

PR #43, opened 2022-10-22
Configurable BucketAuction

me-gozer:
This PR adds the support to enable the bucket auction for a duration whose start and end times are specified with parameters to the constructor. Removed activeStage; and the functions relies on the activeStage are now checking the bucket.timestamp to find out the active stage.

Useful information on the contract changes: 
1. BucketAuction constructor takes two new parameters for the start and end time of the auction.
2. The private property activeStage and its getters/setters were removed. The dependent functions derive this info by checking the bucket.timestamp. 
3. The modifier cannotMint were removed from the BucketAuction::bid and setPrice.
4. From now on the setPrice can not be called after the first token is sent; added a private property and modifier for it.
5. The function ERC721M:canMint assumes the stageTimestamp is block.timestamp unless a timestamp and valid cosigner were specified.

Useful information on the script changes:
1. Created the common/constants to be used by all scripts
2. Added a new script to deploy the bucket auction: deployBA

Useful information on the test changes:
1. Updated the existing test cases affected by the contract changes.
2. Added new test cases to cover the new contract changes.

Test Coverage:

 contracts/                |    98.98 |    88.56 |    97.47 |    98.51 |                |
  BucketAuction.sol        |    98.41 |    88.75 |    96.55 |    97.67 |          62,63 |
  ERC721M.sol              |    98.77 |     90.2 |    97.22 |    98.28 |          58,59 |
All files                  |    97.62 |    87.92 |    96.55 |    97.54 |                |


contracts/ERC721M.sol
```diff
@@ -17,15 +17,15 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
...
+    uint256 internal _activeStage;
     uint256 private _maxMintableSupply;
     uint256 private _globalWalletLimit;
     string private _tokenURISuffix;
     bool private _baseURIPermanent;
     address private _cosigner;
     address private _crossmintAddress;
 
-    MintStageInfo[] private _mintStages;
+    MintStageInfo[] internal _mintStages;
```
nothing0012, line 28: same here, prefer getter function so that we don't leak private access to other contracts

contracts/ERC721M.sol
```diff
@@ -17,15 +17,15 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
 
     bool private _mintable;
     string private _currentBaseURI;
-    uint256 private _activeStage;
+    uint256 internal _activeStage;
```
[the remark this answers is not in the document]
me-gozer: I replaced the modifier auctionStage with the modifier restrictedStageType taking the parameter SaleType; now all of  the variables are private.  

"_or maybe we don't really need _activeStage if we have universal support from start/end timestamp for all stages and bucket/dutch auctions_" --> I will think its use case thoroughly; and might remove/replace it within the next PR.

contracts/IERC721M.sol
```diff
@@ -26,13 +26,17 @@ interface IERC721M is IERC721AQueryable {
     error WalletStageLimitExceeded();
     error WithdrawFailed();
 
+    // The type of the sale
+    enum SaleType { DirectSale, BucketAuction }
```
nothing0012, line 30: this is a leaky abstraction from the more generic contract - I think in general, we probably want the ERC721M to be the base to all the extension contracts, such that it's composable instead of inheritance to define the possible values of all the sale type.

- it adds complexity to the vanila smart contract that now we need to specify saleType for all stages
- adding new sale type requiring bumping the base smart contract npm version. which better to be done purely in the extension contract.

Wondering if we can just change BucketAuction.sol for this PR without adding the saleType, so that the change of determining if bucketAuction is live or not is self contained - i.e. we replace setActive with start/end time in BucketAuction.sol

channing-magiceden:
Can you run coverage and paste the coverage report in PR description?

contracts/ERC721M.sol
```diff
@@ -426,7 +415,7 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
         ) revert InvalidCosignSignature();
     }
 
-    function getActiveStageFromTimestamp(uint64 timestamp)
+    function getActiveStageFromTimestamp(uint256 timestamp)
```
mi-yu, line 418: is this change intentional?

contracts/BucketAuction.sol
```diff
@@ -11,10 +11,13 @@ import "./IBucketAuction.sol";
...
 
 contract BucketAuction is IBucketAuction, ERC721M {
+    
+    uint256 private _startTimeUnixSeconds;
+    uint256 private _endTimeUnixSeconds;
     bool private _claimable;
     uint256 private _minimumContributionInWei;
     uint256 private _price;
-    bool private _auctionActive;
+    bool private _firstTokenSent;
```
mi-yu, line 20: nit: we can group the `bool` variables together at the beginning to save 1 storage slot:

```
bool private _claimable;
bool private _firstTokenSent;
// rest of the variables
```

This is because each storage slot is 32 bytes and fixed size types (bool, uint, address) can be packed into a single storage slot as long as they fit: https://fravoll.github.io/solidity-patterns/tight_variable_packing.html

contracts/BucketAuction.sol
```diff
@@ -24,7 +27,10 @@ contract BucketAuction is IBucketAuction, ERC721M {
         uint256 maxMintableSupply,
         uint256 globalWalletLimit,
         address cosigner,
-        uint256 minimumContributionInWei
+        uint256 minimumContributionInWei,
+        uint256 startTimeUnixSeconds,
```
mi-yu, line 31: these timestamps can be u64 to save a storage slot

contracts/BucketAuction.sol
```diff
@@ -11,10 +11,13 @@ import "./IBucketAuction.sol";
 import "./ERC721M.sol";
 
 contract BucketAuction is IBucketAuction, ERC721M {
+    
+    uint256 private _startTimeUnixSeconds;
```
mi-yu, line 15: these timestamps can be u64 to save a storage slot

scripts/common/constants.ts
```diff
@@ -0,0 +1,6 @@
+// This module wraps up the constants which can be used by any script
+
+export const ContractDetails = {
```
mi-yu, line 3: nice!

contracts/BucketAuction.sol
```diff
@@ -80,21 +97,30 @@ contract BucketAuction is IBucketAuction, ERC721M {
...
     /**
-     * @notice begin the auction.
+     * @notice set the start and end times in unix seconds for the bucket auction.
      * @dev cannot be reactivated after price has been set.
-     * @param b set 'true' to start auction, set 'false' to stop auction.
+     * @param startTime set to unix timestamp for the auction start time.
+     * @param endTime set to unix timestamp for the auction end time.
      */
-    function setAuctionActive(bool b) external onlyOwner cannotMint {
+    function setStartAndEndTimeUnixSeconds(uint256 startTime, uint256 endTime) external onlyOwner {
```
mi-yu, line 105: might also want to check that these times don't overlap with any of the existing stages, and maybe enforce some time gap between stages (~5 mins)

contracts/ERC721M.sol
me-gozer: Yes; it was intentional. 
Now in some code paths bucket.timestamp is passed to the getActiveStageFromTimestamp and it is uint256. Is it too costly to use uint256 instead of uint64? I can cast bucket.timestamp down to uint64.

contracts/BucketAuction.sol
me-gozer: why were we using u256 for _activeStage? Was it intentional?

contracts/ERC721M.sol
me-gozer: FYI I updated its type to uint64 and I will update the PR soon.

mi-yu (approved):

channing-magiceden (approved):

PR #45, opened 2022-10-27
Make timestamp expiry seconds configurable

mi-yu:
On mainnet, 60 seconds is too aggressive causing many transactions to fail with `TimestampExpired`. This PR makes the expiry threshold configurable via:

```
    function setTimestampExpirySeconds(uint64 expiry) external onlyOwner {
        _timestampExpirySeconds = expiry;
        emit SetTimestampExpirySeconds(expiry);
    }
```

Coverage:
```
 contracts/                |      100 |    89.58 |      100 |      100 |                |
  BucketAuction.sol        |      100 |    91.03 |      100 |      100 |                |
  ERC721M.sol              |      100 |    90.74 |      100 |      100 |                |
All files                  |    98.56 |    88.93 |    98.88 |    98.94 |                |
```


contracts/ERC721M.sol
```diff
@@ -443,11 +448,13 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
...
+    function getTimestampExpirySeconds() public view override returns (uint64) {
+        return _timestampExpirySeconds;
+    }
+
     function _assertValidTimestamp(uint64 timestamp) internal view {
-        uint64 threshold = msg.sender == _crossmintAddress
-            ? CROSSMINT_TIMESTAMP_EXPIRY_SECONDS
-            : MIN_STAGE_INTERVAL_SECONDS;
-        if (timestamp < block.timestamp - threshold) revert TimestampExpired();
+        if (timestamp < block.timestamp - getTimestampExpirySeconds())
```
channing-magiceden, line 456: Do you think we should have a different expiry for crossmint?

e.g. add 2 mins on top of getTimestampExpirySeconds()

contracts/BucketAuction.sol
```diff
@@ -32,7 +32,8 @@ contract BucketAuction is IBucketAuction, ERC721M {
             tokenURISuffix,
             maxMintableSupply,
             globalWalletLimit,
-            cosigner
+            cosigner,
+            300
```
channing-magiceden, line 36: nit: add a parameter comment
```
/* timestampExpirySeconds= */ 300,
```

Also, is 5 min too long?

channing-magiceden:
Pls also add test coverage in description.

contracts/ERC721M.sol
mi-yu: I think it will be fine, as long as we pick a large enough expiry. That way we save some gas since we don't have to check `msg.sender` or store a separate expiry variable for crossmint

contracts/BucketAuction.sol
mi-yu: sounds good will add, UX wise i think its fine and it can always be updated later

nothing0012 (approved):
lgtm

PR #51, opened 2022-10-31
Update deploy script for the new constructor param

channing-magiceden:
Update deploy script for the new param `timestampexpiryseconds`

### DIFF
--- a/scripts/deploy.ts
+++ b/scripts/deploy.ts
@@ -13,7 +13,7 @@ export interface IDeployParams {
   maxsupply: string;
   globalwalletlimit: string;
   cosigner?: string;
-  timestampexpiryseconds: number;
+  timestampexpiryseconds?: number;
 }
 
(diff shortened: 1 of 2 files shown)


scripts/deploy.ts
```diff
@@ -35,6 +36,7 @@ export const deploy = async (
     hre.ethers.BigNumber.from(args.maxsupply),
     hre.ethers.BigNumber.from(args.globalwalletlimit),
     args.cosigner ?? hre.ethers.constants.AddressZero,
+    args.timestampexpiryseconds,
```
mi-yu, line 39: maybe we can provide a default of 300 seconds here

mi-yu (approved):

scripts/deploy.ts
channing-magiceden: Updated

PR #53, opened 2022-11-01
Add method to read pages of bids

mi-yu:
- adds `getUserDataPage(uint256 limit, uint256 offset)` which can be used to page through the bids
- use `EnumerableSet` to track unique bidder addresses
- note that the order is not guaranteed to be stable for `EnumerableSet`, so `page(offset=0,limit=0)` may return different data as more bids are added over time

## Coverage
```
 contracts/                |    99.04 |    88.24 |    97.59 |    98.61 |                |
  BucketAuction.sol        |    98.63 |    89.02 |    96.77 |       98 |          69,73 |
  ERC721M.sol              |     98.8 |    89.22 |    97.37 |    98.35 |          58,59 |
All files                  |    97.75 |     87.6 |     96.7 |    97.69 |                |
```

### DIFF
--- a/contracts/BucketAuction.sol
+++ b/contracts/BucketAuction.sol
@@ -5,12 +5,15 @@ pragma solidity ^0.8.4;
 import "@openzeppelin/contracts/access/AccessControl.sol";
 import "@openzeppelin/contracts/access/Ownable.sol";
 import "@openzeppelin/contracts/token/common/ERC2981.sol";
+import "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
 import "erc721a/contracts/ERC721A.sol";
 
 import "./IBucketAuction.sol";
 import "./ERC721M.sol";
 
 contract BucketAuction is IBucketAuction, ERC721M {
+    using EnumerableSet for EnumerableSet.AddressSet;
+
     bool private _claimable;
     bool private _firstTokenSent;
     uint64 private _startTimeUnixSeconds;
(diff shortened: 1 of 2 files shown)


contracts/BucketAuction.sol
```diff
@@ -95,6 +99,37 @@ contract BucketAuction is IBucketAuction, ERC721M {
...
+        view
+        returns (
+            User[] memory,
+            address[] memory,
+            uint256 total
+        )
+    {
+        uint256 numUsers = _users.length();
+        uint256 pageSize = limit;
+        if (pageSize > numUsers - offset) {
```
channing-magiceden, line 113: could be `numUsers - offset` underflow?

channing-magiceden: If offset > numUsers, we should simply return empty.

contracts/BucketAuction.sol
```diff
@@ -95,6 +99,37 @@ contract BucketAuction is IBucketAuction, ERC721M {
...
+        external
+        view
+        returns (
+            User[] memory,
+            address[] memory,
+            uint256 total
+        )
+    {
+        uint256 numUsers = _users.length();
+        uint256 pageSize = limit;
```
channing-magiceden, line 112: Do we need to cap `limit`? For example, maximum 100 etc.

contracts/BucketAuction.sol
mi-yu: i think its fine to let it underflow since this is just a view function, the caller can be responsible for providing valid inputs

contracts/BucketAuction.sol
mi-yu: IMO its ok to not have limits here since the RPC node processing the request will just throw an error if the requested page limit is too expensive. Since the limit might be different across providers/depending on load, we can just make it configurable on the client side

channing-magiceden (approved):
Discussed offline with Rex and AE team. Good to go!

PR #61, opened 2022-11-14
Add setTimestampExpirySeconds script

mi-yu:


channing-magiceden (approved):

channing-magiceden:
I think it worth adding a comment saying updating should be careful as you may need to update stage as well.

PR #65, opened 2022-11-20
Build ESM module for npm

mi-yu:
Export both commonjs and ES module in npm bundle. ES module can help reduce javascript bundle sizes for users of `@magiceden-oss/erc721m` with its improved tree-shaking capabilities. Roughly follows Looksrare's SDK bundling config: https://github.com/LooksRare/looksrare-sdk/blob/master/rollup.config.js

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,12 +8,14 @@
     "!/contracts/mocks/**/*",
     "/dist/**/*"
   ],
-  "main": "dist/src/index.js",
+  "main": "dist/index.cjs.js",
+  "module": "dist/index.esm.js",
+  "types": "dist/src/index.d.ts",
   "scripts": {
     "node": "hardhat node",
     "build": "npm run build:contracts && npm run build:typescript",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -8,12 +8,18 @@
...
     "/dist/**/*"
   ],
-  "main": "dist/src/index.js",
+  "main": "dist/index.cjs.js",
+  "module": "dist/index.esm.js",
+  "types": "dist/src/index.d.ts",
+  "exports": {
+    "require": "./dist/index.cjs.js",
+    "import": "./dist/index.esm.js"
+  },
```
nothing0012, line 17: that works for both?

nothing0012 (approved):
lgtm!

package.json
mi-yu: Yep uses conditional exports so it can be consumed as either esm or commonjs: https://nodejs.org/api/packages.html#conditional-exports

PR #70, opened 2023-01-04
Add operator filtering

channing-magiceden:
- Adopt a light version of OS operator filterer: https://github.com/ProjectOpenSea/operator-filter-registry
- Verified on OS goerli
image

### DIFF
--- a/contracts/OperatorFilter/DefaultOperatorFilterer.sol
+++ b/contracts/OperatorFilter/DefaultOperatorFilterer.sol
@@ -1,0 +1,11 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.13;
+
+import "./OperatorFilterer.sol";
+
+abstract contract DefaultOperatorFilterer is OperatorFilterer {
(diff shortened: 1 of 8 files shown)


contracts/OperatorFilter/DefaultOperatorFilterer.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.13;
```
mi-yu, line 2: any reason not to use the npm package https://www.npmjs.com/package/operator-filter-registry?

mi-yu (approved):

contracts/OperatorFilter/DefaultOperatorFilterer.sol
channing-magiceden: This is based on your previous diff: https://github.com/magiceden-oss/erc721m/pull/70/commits/62e3af54853d93ef85c5a23d7cd5b863d0089f9e

I forgot the reason to have our own. The code seems simpler than the OS version.

PR #80, opened 2023-04-20
 Auto-approve ME conduit on mint

destinymagicchen:
1. Create a new contract extending [ERC721MOperatorFilterer](https://github.com/magiceden-oss/erc721m/blob/main/contracts/ERC721MOperatorFilterer.sol) that takes an additional argument autoApproveAddress. We may also want to have another version that extends the vanilla ERC721M contract in case we don't need the OperatorFilter.
2. Override the mint function to call setApprovalForAll(autoApproveAddress) after super.mint(...). We can also try some gas optimization by first checking isApprovedForAll since reads are cheaper than writes
3. Update the deploy script to handle this new variant https://github.com/magiceden-oss/erc721m/blob/main/scripts/deploy.ts#L22
4. Update tests https://github.com/magiceden-oss/erc721m/tree/main/test, best to create a new file as the existing test files are already super long

### DIFF
--- a/contracts/ERC721MAutoApprover.sol
+++ b/contracts/ERC721MAutoApprover.sol
@@ -6,6 +6,8 @@ import "./ERC721M.sol";
 
 contract ERC721MAutoApprover is ERC721M {
     address private _autoApproveAddress;
+    
+    event SetAutoApproveAddress(address autoApproveAddress);
 
     constructor(
         string memory collectionName,
@@ -46,5 +48,16 @@ contract ERC721MAutoApprover is ERC721M {
             // approve the address if not already approved
             super.setApprovalForAll(_autoApproveAddress, true);
         }
+    }
+
+    function getAutoApproveAddress() external view returns (address) {
+        return _autoApproveAddress;
+    }
+
+    function setAutoApproveAddress(
+        address autoApproveAddress
+    ) external onlyOwner {
+        _autoApproveAddress = autoApproveAddress;
+        emit SetAutoApproveAddress(autoApproveAddress);
     }
 }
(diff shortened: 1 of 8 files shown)


contracts/ERC721MAutoApprover.sol
```diff
@@ -0,0 +1,52 @@
...
+        bytes calldata signature
+    ) external payable override nonReentrant {
+        _mintInternal(qty, msg.sender, proof, timestamp, signature);
+
+        // if auto approve address is not all zero, check if the address is already approved
+        if (
+            _autoApproveAddress != address(0) &&
+            !super.isApprovedForAll(msg.sender, _autoApproveAddress)
+        ) {
+            if (!super.isApprovedForAll(msg.sender, _autoApproveAddress)) {
```
nothing0012, line 46: do we need this line?

contracts/ERC721MOperatorFiltererAutoApprover.sol
```diff
@@ -0,0 +1,52 @@
...
+        bytes calldata signature
+    ) external payable override nonReentrant {
+        _mintInternal(qty, msg.sender, proof, timestamp, signature);
+
+        // if auto approve address is not all zero, check if the address is already approved
+        if (
+            _autoApproveAddress != address(0) &&
+            !super.isApprovedForAll(msg.sender, _autoApproveAddress)
+        ) {
+            if (!super.isApprovedForAll(msg.sender, _autoApproveAddress)) {
```
nothing0012, line 46: same here

nothing0012 (approved):

contracts/ERC721MOperatorFiltererAutoApprover.sol
```diff
@@ -0,0 +1,50 @@
...
+
+        // if auto approve address is not all zero, check if the address is already approved
+        if (
+            _autoApproveAddress != address(0) &&
+            !super.isApprovedForAll(msg.sender, _autoApproveAddress)
+        ) {
+            // approve the address if not already approved
+            super.setApprovalForAll(_autoApproveAddress, true);
+        }
+    }
```
mi-yu, line 49: we should also add functions for `getAutoApproveAddress` and `setAutoApproveAddress` (in case we want to turn off the behavior)

mi-yu (approved):

PR #87, opened 2023-05-24
Pause

ricorandom:
Added the pause feature and the import for the ownable contract.

### DIFF
--- a/contracts/ERC721MOperatorFilterer.sol
+++ b/contracts/ERC721MOperatorFilterer.sol
@@ -6,6 +6,8 @@ import "./ERC721M.sol";
 import "./OperatorFilter/DefaultOperatorFilterer.sol";
 
 contract ERC721MOperatorFilterer is ERC721M, DefaultOperatorFilterer {
+    bool public _tradable = false;
+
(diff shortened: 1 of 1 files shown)


contracts/ERC721MOperatorFilterer.sol
```diff
@@ -50,4 +54,22 @@ contract ERC721MOperatorFilterer is ERC721M, DefaultOperatorFilterer {
...
+
+    function _beforeTokenTransfer(
+        address from,
+        address to,
+        uint256 tokenId
+    ) internal virtual override(ERC721) {
+        if (
+            _tradable != true
+        ) {
+            revert("Transfers are currently disabled");
```
mi-yu, line 71: prefer to revert an error: https://github.com/magicoss/erc721m/blob/main/contracts/IERC721M.sol#L7-L27

ricorandom:
Hi Mi Yu,

Pausable <https://docs.openzeppelin.com/contracts/4.x/api/security#Pausable>
works, i'd be happy to use that instead, do you want me to make that
amendment?

PR #88, opened 2023-06-21
Add BucketAuctionOperatorFilterer

channing-magiceden:
Add support of BucketAuction with OperatorFilterer.


mi-yu:
The lock file changes seem very large, is it expected? Also we should only have package-lock.json as we're just using npm for package manager

Henry-MagicEden (dismissed):

channing-magiceden:
> 

You are right. I accidentally added `yarn.lock`. Removed.

channing-magiceden:
> The lock file changes seem very large, is it expected? Also we should only have package-lock.json as we're just using npm for package manager


You are right. I accidentally added yarn.lock. Removed.

channing-magiceden:
Removed new package to unblock

mi-yu (approved):

PR #90, opened 2023-06-28
initial working erc721mc with tests

jharris:
Combines the vanilla ERC721M contracts with the Limit Break "creator token" and programmable royalties extensions.
TODO: additional hardhat tests need to be ported over from Limit Break's solidity-based tests.


jharris:
Oops, didn't mean to open this on ME's github. Pls Ignore for now.

PR #94, opened 2023-08-01
Minting with ERC-20 payment

Henry-MagicEden:
This adds the ability to set the ERC721M contract to accept payment in an ERC-20 token rather than the native currency. These changes are non-breaking and should allow the contract to operate as before if native currency payment is all that's needed.

- `mintCurrency` is the contract address of the ERC-20 token to accept payment in. If set to 0 address (default), the contract accepts the native currency for minting
- Using OZ's [SafeERC20](https://docs.openzeppelin.com/contracts/2.x/api/token/erc20#SafeERC20) for its `safeTransferFrom` and `safeTransfer` functions that throw on failure

------

 - Unit tests complete
 - Next step is testing this manually on testnet

### DIFF
--- a/contracts/ERC721M.sol
+++ b/contracts/ERC721M.sol
@@ -375,16 +375,8 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
     /**
      * @dev Returns mint currency address.
      */
-    function getMintCurrency() external view returns (address) {
+    function getMintCurrency() external view override returns (address) {
         return _mintCurrency;
-    }
-
-    /**
-     * @dev Sets the mint currency used for payment.
-     */
-    function setMintCurrency(address mintCurrency) external onlyOwner {
-        _mintCurrency = mintCurrency;
-        emit SetMintCurrency(mintCurrency);
     }
 
     /**
@@ -520,10 +512,10 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
     /**
      * @dev Withdraws ERC-20 funds by owner.
      */
-    function withdrawERC20(address mintCurrency) external onlyOwner {
-        uint256 value = IERC20(mintCurrency).balanceOf(address(this));
-        IERC20(mintCurrency).safeTransfer(msg.sender, value);
-        emit WithdrawERC20(mintCurrency, value);
+    function withdrawERC20() external onlyOwner {
+        uint256 value = IERC20(_mintCurrency).balanceOf(address(this));
+        IERC20(_mintCurrency).safeTransfer(msg.sender, value);
+        emit WithdrawERC20(_mintCurrency, value);
     }
 
     /**
--- a/contracts/ERC721M.sol
+++ b/contracts/ERC721M.sol
@@ -513,6 +513,7 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
      * @dev Withdraws ERC-20 funds by owner.
      */
     function withdrawERC20() external onlyOwner {
+        if (_mintCurrency == address(0)) revert WrongMintCurrency();
         uint256 value = IERC20(_mintCurrency).balanceOf(address(this));
         IERC20(_mintCurrency).safeTransfer(msg.sender, value);
         emit WithdrawERC20(_mintCurrency, value);
--- a/contracts/ERC721M.sol
+++ b/contracts/ERC721M.sol
@@ -63,7 +63,7 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
     mapping(uint256 => uint256) private _stageMintedCounts;
 
     // Address of ERC-20 token used to pay for minting. If 0 address, use native currency.
-    address public mintCurrency;
+    address private _mintCurrency;
 
     constructor(
         string memory collectionName,
@@ -72,7 +72,8 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
         uint256 maxMintableSupply,
         uint256 globalWalletLimit,
         address cosigner,
-        uint64 timestampExpirySeconds
+        uint64 timestampExpirySeconds,
+        address mintCurrency
     ) ERC721A(collectionName, collectionSymbol) {
         if (globalWalletLimit > maxMintableSupply)
             revert GlobalWalletLimitOverflow();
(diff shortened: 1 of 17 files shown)


contracts/ERC721M.sol
```diff
@@ -365,6 +371,14 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
         );
     }
 
+    /**
+     * @dev Sets the mint currency used for payment.
+     */
+    function setMintCurrency(address newMintCurrency) external onlyOwner {
```
nothing0012, line 377: let's disable set mint currency, better just set from constructor

contracts/ERC721M.sol
```diff
@@ -491,6 +509,15 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
         emit Withdraw(value);
     }
 
+    /**
+     * @dev Withdraws ERC-20 funds by owner.
+     */
+    function withdrawERC20(address mintCurrencyWithdrawn) external onlyOwner {
```
mi-yu, line 515: thoughts on allowing a recipient address to be passed here?

contracts/ERC721M.sol
Henry-MagicEden: Hmm I think it could make sense to also set from constructor, but what's the reason for removing the setter?

contracts/ERC721M.sol
```diff
@@ -60,6 +62,9 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
     // Minted count per stage.
     mapping(uint256 => uint256) private _stageMintedCounts;
 
+    // Address of ERC-20 token used to pay for minting. If 0 address, use native currency.
+    address public mintCurrency;
```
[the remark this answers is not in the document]
Henry-MagicEden: Our contract is already large, which makes it difficult to integrate with other tools like LayerZero (cc @channing-magiceden) - making it public gives us a getter automatically without another function and doesn't compromise security

Is the concern more around styling? Or is there a functional reason I'm missing here?

contracts/ERC721M.sol
```diff
@@ -491,6 +509,15 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
         emit Withdraw(value);
     }
 
+    /**
+     * @dev Withdraws ERC-20 funds by owner.
+     */
+    function withdrawERC20(address mintCurrencyWithdrawn) external onlyOwner {
+        uint256 value = IERC20(mintCurrencyWithdrawn).balanceOf(address(this));
```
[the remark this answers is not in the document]
Henry-MagicEden: Do we need to? It will just throw right?

contracts/ERC721M.sol
```diff
@@ -462,6 +476,10 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
             ) revert InvalidProof();
         }
 
+        if (mintCurrency != address(0)) {
+            IERC20(mintCurrency).safeTransferFrom(msg.sender, address(this), stage.price * qty);
```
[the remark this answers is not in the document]
Henry-MagicEden: Yep - it will revert if the owner doesn't have enough of the token, or if the spender doesn't have enough approval

contracts/ERC721M.sol
Henry-MagicEden: @nothing0012 is there a reason there isn't an address param in the native currency `withdraw` function? I followed the convention there

If no reason, happy to add the param to this one

contracts/ERC721M.sol
nothing0012: it prevents admin accidentally changing the mint currency and can rug users

contracts/ERC721M.sol
```diff
@@ -491,6 +517,15 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
         emit Withdraw(value);
     }
 
+    /**
+     * @dev Withdraws ERC-20 funds by owner.
+     */
+    function withdrawERC20(address mintCurrency) external onlyOwner {
+        uint256 value = IERC20(mintCurrency).balanceOf(address(this));
```
nothing0012, line 524: _mintCurrency

contracts/ERC721M.sol
```diff
@@ -491,6 +517,15 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
         emit Withdraw(value);
     }
 
+    /**
+     * @dev Withdraws ERC-20 funds by owner.
+     */
+    function withdrawERC20(address mintCurrency) external onlyOwner {
```
nothing0012, line 523: do we need this input? for example, we can directly use _mintCurrency

contracts/ERC721M.sol
```diff
@@ -365,6 +372,21 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
...
+        return _mintCurrency;
+    }
+
+    /**
+     * @dev Sets the mint currency used for payment.
+     */
+    function setMintCurrency(address mintCurrency) external onlyOwner {
+        _mintCurrency = mintCurrency;
+        emit SetMintCurrency(mintCurrency);
+    }
```
nothing0012, line 388: thinking about setting mintCurrency again, is there a need for it to be per stage?

contracts/ERC721M.sol
Henry-MagicEden: Yea, this was while there was still a setter, so I guess it can be removed

contracts/ERC721M.sol
Henry-MagicEden: Not for the first version, since we're just testing it out as a feature to see how many projects want it before adding more complexity to it

contracts/ERC721M.sol
mi-yu: i think its not required, but could be nice if we want to distribute funds directly from the contract to the project since it can save us an extra transfer

contracts/ERC721M.sol
```diff
@@ -491,6 +509,15 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
         emit Withdraw(value);
     }
 
+    /**
+     * @dev Withdraws ERC-20 funds by owner.
+     */
+    function withdrawERC20() external onlyOwner {
+        uint256 value = IERC20(_mintCurrency).balanceOf(address(this));
```
nothing0012, line 516: just curious will it revert if _mintCurrency is address(0)? let's unit test this

contracts/ERC721M.sol
Henry-MagicEden: Think either way there's an extra transfer, since the project would have to send us our share so they'd have to do the transfer

Medium-term the plan is actually to consider inheriting from the PaymentSplitter so that ME and the Creator can each just get their own share without having to go through the other

contracts/ERC721M.sol
Henry-MagicEden: Added an explicit revert and new error - previously, it was returning this error `Error: Transaction reverted: function returned an unexpected amount of data` which is not super informative

mi-yu (approved):

PR #95, opened 2023-08-14
Add OperatorFilterer into ERC721MOnft

channing-magiceden:
- Remove functions not being used in `IERC721M`
- Further downsize `IERC721MLite` by removing crossminting and updating specific stages
- Add OperatorFilterer into `IERC721MLite` 
- Bump package version from `0.0.10` to `0.0.11`

image

### DIFF
--- a/contracts/ERC721M.sol
+++ b/contracts/ERC721M.sol
@@ -114,7 +114,7 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
      */
     function setCosigner(address cosigner) external onlyOwner {
         _cosigner = cosigner;
-        emit SetCosigner(_cosigner);
+        emit SetCosigner(cosigner);
     }
 
     /**
@@ -130,7 +130,7 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
      */
     function setCrossmintAddress(address crossmintAddress) external onlyOwner {
         _crossmintAddress = crossmintAddress;
-        emit SetCrossmintAddress(_crossmintAddress);
+        emit SetCrossmintAddress(crossmintAddress);
     }
 
     /**
--- a/contracts/ERC721M.sol
+++ b/contracts/ERC721M.sol
@@ -95,26 +95,11 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
     }
 
     /**
-     * @dev Returns whether NOT mintable.
-     */
-    modifier cannotMint() {
-        if (_mintable) revert Mintable();
-        _;
-    }
-
-    /**
      * @dev Returns whether it has enough supply for the given qty.
      */
     modifier hasSupply(uint256 qty) {
         if (totalSupply() + qty > _maxMintableSupply) revert NoSupplyLeft();
         _;
-    }
-
-    /**
-     * @dev Returns cosigner address.
-     */
-    function getCosigner() external view override returns (address) {
-        return _cosigner;
     }
 
     /**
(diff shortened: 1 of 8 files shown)


mi-yu (approved):

contracts/ERC721M.sol
```diff
@@ -110,13 +102,6 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
         _;
     }
 
-    /**
-     * @dev Returns cosigner address.
-     */
-    function getCosigner() external view override returns (address) {
```
Henry-MagicEden, line 116: Do these need to be removed from the main `ERC721M` contract? Could we just remove them from IERC721M but keep them in the actual contract?

channing-magiceden: It's not quite useful and ME has no references. We don't have any use case that could change cosigner. And we don't have issues historically to mis-set cosigner. Maybe good to keep them for debugging convenience but I think it's easier to just reset `cosigner`.

contracts/ERC721M.sol
```diff
@@ -94,14 +94,6 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
         _;
     }
 
-    /**
-     * @dev Returns whether NOT mintable.
-     */
-    modifier cannotMint() {
```
nothing0012, line 100: why remove this? not used anymore?

contracts/ERC721M.sol
```diff
@@ -129,14 +114,7 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
...
         _cosigner = cosigner;
-        emit SetCosigner(cosigner);
-    }
-
-    /**
-     * @dev Returns expiry in seconds.
-     */
-    function getTimestampExpirySeconds() public view override returns (uint64) {
-        return _timestampExpirySeconds;
+        emit SetCosigner(_cosigner);
```
nothing0012, line 117: let's use cosigner to avoid a storage read

contracts/ERC721M.sol
```diff
@@ -535,18 +505,6 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
         emit PermanentBaseURI(_currentBaseURI);
     }
 
-    /**
-     * @dev Returns token URI suffix.
-     */
-    function getTokenURISuffix()
```
nothing0012, line 541: we need this, right?

contracts/ERC721M.sol
```diff
@@ -147,19 +125,12 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
...
-        return _crossmintAddress;
-    }
-
     /**
      * @dev Sets crossmint address if using crossmint. This allows the specified address to call `crossmint`.
      */
     function setCrossmintAddress(address crossmintAddress) external onlyOwner {
         _crossmintAddress = crossmintAddress;
-        emit SetCrossmintAddress(crossmintAddress);
+        emit SetCrossmintAddress(_crossmintAddress);
```
nothing0012, line 133: this uses more gas, is it really going to save the binary size?

contracts/ERC721M.sol
qianmao: Yeah not used.

channing-magiceden: Yeah not used.

contracts/ERC721M.sol
channing-magiceden: Not quite, we never use this externally.

nothing0012 (approved):

PR #102, opened 2023-12-22
feat: add mint fee

channing-magiceden:
- Besides price, add optional mint fee
- Mint fee goes to dedicated address when `withdraw` or `withdrawERC20` is invoked
- Support both ETH and ERC20 payment


contracts/ERC721M.sol
```diff
@@ -65,6 +65,9 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
     // Address of ERC-20 token used to pay for minting. If 0 address, use native currency.
     address private _mintCurrency;
 
+    address public constant MINT_FEE_RECEIVER =
+        address(0x0B98151bEdeE73f9Ba5F2C7b72dEa02D38Ce49Fc);
```
wolfy-nft, line 69: we can get some gas savings by making this immutable since we don't have any update logic for it.

```
    address public immutable MINT_FEE_RECEIVER;

    constructor() {
        MINT_FEE_RECEIVER = address(0x0B98151bEdeE73f9Ba5F2C7b72dEa02D38Ce49Fc);
    }
```

wolfy-nft: also we can remove the address() casting since its already an address

contracts/ERC721M.sol
wolfy-nft, line 1:
```
    function setStages(MintStageInfo[] calldata newStages) external onlyOwner {
        uint256 originalSize = _mintStages.length;
        for (uint256 i = 0; i < originalSize; i++) {
            _mintStages.pop();
        }
```

this can be replaced with:
```
delete _mintStages
```
also major gas savings

contracts/ERC721M.sol
```diff
@@ -448,11 +457,23 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
         }
 
         if (_mintCurrency != address(0)) {
+            // ERC20 mint payment
```
wolfy-nft, line 460: why don't we manually transfer to mint fee receiver after the mint has ended? transferring to the receiver for every mint adds a lot of extra gas for the user (wasted $$$)

contracts/ERC721M.sol
```diff
@@ -448,11 +457,23 @@ contract ERC721M is IERC721M, ERC721AQueryable, Ownable, ReentrancyGuard {
...
             IERC20(_mintCurrency).safeTransferFrom(
```
wolfy-nft, line 461: we should just use transferFrom for both calls.

safeTransferFrom is meant as a check for smart contracts that can support erc20 token transfers. Since we're a launchpad i think its fair that we shouldnt support smart contracts minting from our launchpad, and shouldnt care if they fail to accept erc20 tokens.

for MINT_FEE_RECEIVER, if its a contract we own then we can safely not use safeTransferFrom as long as we know it can accept erc20 tokens. if its a EOA address we should just use transferFrom.

calling safeTransferFrom also has extra gas over transferFrom

contracts/ERC721M.sol
channing-magiceden: Good point. Updated.

contracts/ERC721M.sol
channing-magiceden: +1 agree. Fixed.

contracts/ERC721M.sol
channing-magiceden: Actually. `const` is better than `immutable` as it's compiled into bytecode thus has lower gas cost. 

For the suggestion of removing casting, agreed and fixed!

contracts/ERC721M.sol
channing-magiceden: Fixed.

contracts/ERC721CM.sol
```diff
@@ -66,6 +66,12 @@ contract ERC721CM is IERC721M, ERC721ACQueryable, Ownable, ReentrancyGuard {
     // Address of ERC-20 token used to pay for minting. If 0 address, use native currency.
     address private _mintCurrency;
 
+    // Total mint fee
+    uint256 private totalMintFee;
+
+    address public constant MINT_FEE_RECEIVER =
+        address(0x0B98151bEdeE73f9Ba5F2C7b72dEa02D38Ce49Fc);
```
wolfy-nft, line 73: casting address here is redundant

contracts/ERC721CM.sol
```diff
@@ -66,6 +66,12 @@ contract ERC721CM is IERC721M, ERC721ACQueryable, Ownable, ReentrancyGuard {
     // Address of ERC-20 token used to pay for minting. If 0 address, use native currency.
     address private _mintCurrency;
 
+    // Total mint fee
+    uint256 private totalMintFee;
```
wolfy-nft, line 70: should be underscore to be consistent with other private variables

contracts/ERC721CM.sol
```diff
@@ -187,14 +194,17 @@ contract ERC721CM is IERC721M, ERC721ACQueryable, Ownable, ReentrancyGuard {
             emit UpdateStage(
                 i,
                 newStages[i].price,
+                newStages[i].mintFee,
                 newStages[i].walletLimit,
                 newStages[i].merkleRoot,
                 newStages[i].maxStageSupply,
                 newStages[i].startTimeUnixSeconds,
```
wolfy-nft, line 201: nit pick, but i think splitting into a variable might be more gas efficient here
`MintStageInfo calldata stageInfo = newStages[i];`

contracts/ERC721CM.sol
```diff
@@ -160,7 +166,7 @@ contract ERC721CM is IERC721M, ERC721ACQueryable, Ownable, ReentrancyGuard {
     function setStages(MintStageInfo[] calldata newStages) external onlyOwner {
         delete _mintStages;
 
-        for (uint256 i = 0; i < newStages.length;) {
+        for (uint256 i = 0; i < newStages.length; ) {
```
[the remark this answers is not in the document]
channing-magiceden: Yeah it's added by linter so I would keep it.

contracts/ERC721CM.sol
channing-magiceden: Not sure if the compiler will optimize. But I think it's ok as we only call setStages once.

contracts/ERC721CM.sol
```diff
@@ -409,19 +421,24 @@ contract ERC721CM is IERC721M, ERC721ACQueryable, Ownable, ReentrancyGuard {
...
+            if (
+                limit > 0 &&
+                _stageMintedCountsPerWallet[activeStage][to] + qty > limit
+            ) {
                 revert WalletStageLimitExceeded();
             }
         }
 
         if (_mintCurrency != address(0)) {
-            IERC20(_mintCurrency).safeTransferFrom(
```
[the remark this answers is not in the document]
channing-magiceden: OK I think we should add safe back. `SafeERC20.safeTransfer` is not checking receivable but to ensure ERC20 transfer reverts if an issue happens.

wolfy-nft (approved):

PR #107, opened 2024-02-14
Add ERC721BatchTransfer and couple improvements

channing-magiceden:
- Add `ERC721BatchTransfer` to enable 721 token airdrop
  - Mainnet `0x38F7ba911f7efc434D29D6E39c814E9d4De3FEef`
  - Polygon `0x38F7ba911f7efc434D29D6E39c814E9d4De3FEef`
- Add gas estimate for expensive operations
- Refactor scripts a little bit

### DIFF
--- a/contracts/utils/ERC721BatchTransfer.sol
+++ b/contracts/utils/ERC721BatchTransfer.sol
@@ -1,0 +1,159 @@
+//SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.4;
+
+import { IERC721 } from "@openzeppelin/contracts/token/ERC721/IERC721.sol";
+
+/**
+ * @title ERC721 Batch Transfer
+ *
+ * @notice ETransfer ERC721 tokens in batches to a single wallet or multiple wallets. This supports ERC721M and ERC721CM contracts.
+ * @notice To use any of the methods in this contract the user has to approve this contract to control their tokens using either `setApproveForAll` or `approve` functions from the ERC721 contract.
+ */
+contract ERC721BatchTransfer {
+    error InvalidArguments();
+    error NotOwnerOfToken();
+    error InvalidCaller();
+
+    event BatchTransferToSingle(
+        address indexed contractAddress,
+        address indexed to,
+        uint256 amount
+    );
+
+    event BatchTransferToMultiple(
+        address indexed contractAddress,
+        uint256 amount
+    );
+
+    constructor() {}
+
+    modifier noZero() {
+        if (msg.sender == address(0)) revert InvalidCaller();
+        _;
+    }
+
+    /** 
+     * @notice Transfer multiple tokens to the same wallet using the ERC721.transferFrom method.
+     * @notice If you don't know what that means, use the `safeBatchTransferToSingleWallet` method instead
+     * @param erc721Contract the address of the nft contract
+     * @param to the address that will receive the nfts
+     * @param tokenIds the list of tokens that will be transferred
+     */
+    function batchTransferToSingleWallet(
+        IERC721 erc721Contract,
+        address to,
+        uint256[] calldata tokenIds
+    ) external noZero {
+        uint256 length = tokenIds.length;
+        for (uint256 i; i < length; ) {
+            uint256 tokenId = tokenIds[i];
+            address owner = erc721Contract.ownerOf(tokenId);
+            if (msg.sender != owner) {
+                revert NotOwnerOfToken();
+            }
+            erc721Contract.transferFrom(owner, to, tokenId);
+            unchecked {
+                ++i;
+            }
+        }
+        emit BatchTransferToSingle(address(erc721Contract), to, length);
+    }
+
+    /**
+     * @notice transfer multiple tokens to the same wallet using the `ERC721.safeTransferFrom` method
+     * @param erc721Contract the address of the nft contract
+     * @param to the address that will receive the nfts
+     * @param tokenIds the list of tokens that will be transferred
+     */
+    function safeBatchTransferToSingleWallet(
+        IERC721 erc721Contract,
+        address to,
+        uint256[] calldata tokenIds
+    ) external noZero {
+        uint256 length = tokenIds.length;
+        for (uint256 i; i < length; ) {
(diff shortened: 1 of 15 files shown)


contracts/utils/ERC721BatchTransfer.sol
```diff
@@ -0,0 +1,159 @@
+//SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.4;
+
+import { IERC721 } from "@openzeppelin/contracts/token/ERC721/IERC721.sol";
+
+/**
+ * @title ERC721 Batch Transfer
+ *
+ * @notice ETransfer ERC721 tokens in batches to a single wallet or multiple wallets. This supports ERC721M and ERC721CM contracts.
```
wolfy-nft, line 10: minor typo `ETransfer`

contracts/utils/ERC721BatchTransfer.sol
```diff
@@ -0,0 +1,159 @@
...
+
+    event BatchTransferToMultiple(
+        address indexed contractAddress,
+        uint256 amount
+    );
+
+    constructor() {}
+
+    modifier noZero() {
+        if (msg.sender == address(0)) revert InvalidCaller();
```
wolfy-nft, line 32: what is the purpose of this? how could msg.sender ever be the null address?

contracts/utils/ERC721BatchTransfer.sol
```diff
@@ -0,0 +1,159 @@
...
+        address indexed to,
+        uint256 amount
+    );
+
+    event BatchTransferToMultiple(
+        address indexed contractAddress,
+        uint256 amount
+    );
+
+    constructor() {}
```
wolfy-nft, line 29: safe to remove this

contracts/utils/ERC721BatchTransfer.sol
```diff
@@ -0,0 +1,159 @@
...
+     * @param erc721Contract the address of the nft contract
+     * @param to the address that will receive the nfts
+     * @param tokenIds the list of tokens that will be transferred
+     */
+    function batchTransferToSingleWallet(
+        IERC721 erc721Contract,
+        address to,
+        uint256[] calldata tokenIds
+    ) external noZero {
+        uint256 length = tokenIds.length;
```
wolfy-nft, line 48: if tokenIds is an empty array, we are still emitting an BatchTransferToSingle event. will this ever be a problem? we might want to revert if its an empty array

contracts/utils/ERC721BatchTransfer.sol
```diff
@@ -0,0 +1,159 @@
...
+     * @param erc721Contract the address of the nft contract
+     * @param to the address that will receive the nfts
+     * @param tokenIds the list of tokens that will be transferred
+     */
+    function safeBatchTransferToSingleWallet(
+        IERC721 erc721Contract,
+        address to,
+        uint256[] calldata tokenIds
+    ) external noZero {
+        uint256 length = tokenIds.length;
```
wolfy-nft, line 74: same case here about zero length array

contracts/utils/ERC721BatchTransfer.sol
```diff
@@ -0,0 +1,159 @@
...
+    /** 
+     * @notice Transfer multiple tokens to the same wallet using the ERC721.transferFrom method.
+     * @notice If you don't know what that means, use the `safeBatchTransferToSingleWallet` method instead
+     * @param erc721Contract the address of the nft contract
+     * @param to the address that will receive the nfts
+     * @param tokenIds the list of tokens that will be transferred
+     */
+    function batchTransferToSingleWallet(
+        IERC721 erc721Contract,
+        address to,
```
wolfy-nft, line 45: should we revert if `to` is a null address?

contracts/utils/ERC721BatchTransfer.sol
```diff
@@ -0,0 +1,159 @@
...
+
+    /**
+     * @notice transfer multiple tokens to the same wallet using the `ERC721.safeTransferFrom` method
+     * @param erc721Contract the address of the nft contract
+     * @param to the address that will receive the nfts
+     * @param tokenIds the list of tokens that will be transferred
+     */
+    function safeBatchTransferToSingleWallet(
+        IERC721 erc721Contract,
+        address to,
```
wolfy-nft, line 71: revert if zero address?

contracts/utils/ERC721BatchTransfer.sol
channing-magiceden: It's not null address. Address(0) is valid address - we just don't know who has / will have the private key. Just a practice to not allow address(0) as it's known to be used as "null".

contracts/utils/ERC721BatchTransfer.sol
channing-magiceden: yeah, I don't think it's a problem. We emit event whenever the function is called. Anyway we don't have use case to listen to events.

contracts/utils/ERC721BatchTransfer.sol
channing-magiceden: same above.

contracts/utils/ERC721BatchTransfer.sol
channing-magiceden: You mean `address(0)`? We should allow it so we can batch burn.

contracts/utils/ERC721BatchTransfer.sol
channing-magiceden: Same

contracts/utils/ERC721BatchTransfer.sol
wolfy-nft: it is a valid address but nobody has the compute power to generate the private key for address(0) which is why it used as the burn address

it currently holds > $200M in assets
https://etherscan.io/address/0x0000000000000000000000000000000000000000

if we are using this internally then it doesnt matter, just calling out that its not a necessary check to have

channing-magiceden: Ok just remove the check to save more gas.

wolfy-nft (approved):

PR #143, opened 2024-07-25
Random Games 

channing-magiceden:
- Overall structure
image

- Customize 721CM by adding a `call` to Random Games proxy contract at the end of minting

### DIFF
--- a/contracts/RandomGames.sol
+++ b/contracts/RandomGames.sol
@@ -1,0 +1,78 @@
+//SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.4;
+
+import {ERC721CM, ERC721ACQueryable, IERC721A} from "./ERC721CM.sol";
+
+/**
+ * @title RandomGamesMinting
+ */
+contract RandomGamesMinting is ERC721CM {
+    address immutable _proxyContract;
+
+    constructor(
+        string memory collectionName,
+        string memory collectionSymbol,
+        string memory tokenURISuffix,
+        uint256 maxMintableSupply,
+        uint256 globalWalletLimit,
+        address cosigner,
+        uint64 timestampExpirySeconds,
+        address mintCurrency,
+        address fundReceiver,
+        address proxyContract
+    )
+        ERC721CM(
+            collectionName,
+            collectionSymbol,
+            tokenURISuffix,
+            maxMintableSupply,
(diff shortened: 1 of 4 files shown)


wolfy-nft (approved):

contracts/RandomGames.sol
```diff
@@ -0,0 +1,78 @@
...
+            timestampExpirySeconds,
+            mintCurrency,
+            fundReceiver
+        )
+    {
+        _proxyContract = proxyContract;
+    }
+
+    function setProxyContract(address newProxyContract) external onlyOwner {
+        _proxyContract = newProxyContract;
```
tenthirtyone, line 41: `require(newProxyContract != address(0), "Proxy contract cannot be zero address");`

If you want to be able to disable the proxy by setting it to the zero address then I would do the recommendation of adding the check in the mint.

contracts/RandomGames.sol
```diff
@@ -0,0 +1,78 @@
...
+            tokenURISuffix,
+            maxMintableSupply,
+            globalWalletLimit,
+            cosigner,
+            timestampExpirySeconds,
+            mintCurrency,
+            fundReceiver
+        )
+    {
+        _proxyContract = proxyContract;
```
tenthirtyone, line 37: `require(newProxyContract != address(0), "Proxy contract cannot be zero address");`

You may want to deploy without setting an address, I see your setter function. Then I would add a similar check to the `mint` and `mintWithLimit` functions to ensure the `_proxyContract` address has been set at the beginning of the funciton.

contracts/RandomGames.sol
```diff
@@ -0,0 +1,78 @@
...
+    function mint(
+        uint32 qty,
+        bytes32[] calldata proof,
+        uint64 timestamp,
+        bytes calldata signature
+    ) external override payable virtual nonReentrant {
+        _mintInternal(qty, msg.sender, 0, proof, timestamp, signature);
+
+        address[] memory minters = new address[](qty);
+        for (uint256 i = 0; i < qty; i++) {
```
tenthirtyone, line 53: It is 20,000 gas to write to an array. So there is a theoretical limit here to how large qty could be before the loop will run out of gas. It will probably in the range of 1,000 - 1,500 on the high end.

PR #146, opened 2024-08-21
Add authorizedMint and whitelist Reservoir Relay EOA

channing-magiceden:
- Add `authorizedMint` in 1155 and 721
- Whitelisted Reservoir Relay EOA by default


tenthirtyone:
If that EOA is compromised or the relationship with the EOA changes I'm not seeing any protections that can ever remove it. I think the `_authorizedMinters` mapping needs add/remove methods and a process for managing access to those functions.

What is the motivation for setting the EoA as a constant if it can be passed as a constructor param?

channing-magiceden:
> If that EOA is compromised or the relationship with the EOA changes I'm not seeing any protections that can ever remove it. I think the `_authorizedMinters` mapping needs add/remove methods and a process for managing access to those functions.
> 
> What is the motivation for setting the EoA as a constant if it can be passed as a constructor param?

Good suggestion. My initial intention is to make Reservior Relay fixed so it cannot be modified. But after some thoughts I think it's  better to make the list mutable (by owner only).

tenthirtyone:
Should work. 

In the future what do you think about something like `AuthorizedMinter.sol` and inherit/extend but it's not necessary in this PR, I think.

tenthirtyone (approved):
See comments about `AuthorizedMinter.sol` in the future. There could probably be some zero address checks but it should be benign/caught by the underlying standard.

channing-magiceden:
> Should work.
> 
> In the future what do you think about something like `AuthorizedMinter.sol` and inherit/extend but it's not necessary in this PR, I think.

I think we could switch to use OZ's Access Control / `Role` to consolidate `owner` and `authorizer`.

nothing0012:
worth refactoring later, but yeah, lgtm for now, that we delay the introduction of another Roles dependency

PR #148, opened 2024-08-26
Update scripts for set stages to handle multiple ids for 1155 mints

raymond-me:
`set1155Stages.ts` was previously only compatible with single-edition 1155 mints. This PR adds the handling for multiple token ids while keeping the script backwards compatible.

Note: source of truth for editions is the `price` field. 3 elements in price field means there are 3 editions.

### DIFF
--- a/scripts/set1155Stages.ts
+++ b/scripts/set1155Stages.ts
@@ -3,7 +3,7 @@ import { HardhatRuntimeEnvironment } from 'hardhat/types';
 import { MerkleTree } from 'merkletreejs';
 import fs from 'fs';
 import { ContractDetails } from './common/constants';
-import { cleanVariableWalletLimit, cleanWhitelist, estimateGas } from './utils/helper';
+import { cleanVariableWalletLimit, cleanWhitelist, ensureArray, estimateGas } from './utils/helper';
 import { Overrides } from 'ethers';
 
 export interface ISet1155StagesParams {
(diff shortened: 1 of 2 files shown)


scripts/set1155Stages.ts
```diff
@@ -91,15 +91,20 @@ export const set1155Stages = async (
...
-    price: [ethers.utils.parseEther(s.price)],
-    mintFee: [s.mintFee ? ethers.utils.parseEther(s.mintFee) : 0],
-    maxStageSupply: [s.maxSupply ?? 0],
-    walletLimit: [s.walletLimit ?? 0],
-    merkleRoot: [merkleRoots[i]],
-    startTimeUnixSeconds: Math.floor(new Date(s.startDate).getTime() / 1000),
-    endTimeUnixSeconds: Math.floor(new Date(s.endDate).getTime() / 1000),
-  }));
+  const stages = stagesConfig.map((s, i) => {
+    const price = ensureArray(s.price).map(p => ethers.utils.parseEther(p));
```
channing-magiceden, line 95: Why `price` is handled differently compared to other fiends such as `mintFee`?

raymond-me: @channing-magiceden 

It's being used as source of truth (elements in price array dictates how many editions). Another more expressive way I just thought to do now is to pass in editions as a variable and validate lengths here instead of using `price`.

channing-magiceden (approved):

channing-magiceden:
Hi @nothing0012 can you take another look and help merge the code? Looks good to me.

nothing0012 (approved):

PR #151, opened 2024-09-06
Add cosigner in 1155M

channing-magiceden:
- Add cosigner into 1155M, similar to 721M
- Signed signature:
 ```
  keccak256(
      abi.encodePacked(
          address(this),
          minter,
          qty,
          _cosigner,
          timestamp,
          block.chainid,
          getCosignNonce(minter, tokenId)
      )
  )
 ```

### DIFF
--- a/contracts/ERC1155M.sol
+++ b/contracts/ERC1155M.sol
@@ -7,6 +7,8 @@ import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
 import "@openzeppelin/contracts/utils/cryptography/MerkleProof.sol";
 import "@openzeppelin/contracts/token/common/ERC2981.sol";
 import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "@openzeppelin/contracts/utils/cryptography/MessageHashUtils.sol";
+import "@openzeppelin/contracts/utils/cryptography/SignatureChecker.sol";
 import "@openzeppelin/contracts/utils/ReentrancyGuard.sol";
 import "./utils/Constants.sol";
 
@@ -55,6 +57,12 @@ contract ERC1155M is IERC1155M, ERC1155Supply, ERC2981, Ownable2Step, Reentrancy
     // Total mint fee
     uint256 private _totalMintFee;
 
+    // Specify how long a signature from cosigner is valid for, recommend 300 seconds.
+    uint64 private _timestampExpirySeconds;
+
+    // The address of the cosigner server.
+    address private _cosigner;
+
     // Address of ERC-20 token used to pay for minting. If 0 address, use native currency.
     address private immutable MINT_CURRENCY;
 
@@ -73,6 +81,8 @@ contract ERC1155M is IERC1155M, ERC1155Supply, ERC2981, Ownable2Step, Reentrancy
         string memory uri,
         uint256[] memory maxMintableSupply,
         uint256[] memory globalWalletLimit,
+        address cosigner,
+        uint64 timestampExpirySeconds,
         address mintCurrency,
         address fundReceiver,
         address royaltyReceiver,
@@ -93,8 +103,11 @@ contract ERC1155M is IERC1155M, ERC1155Supply, ERC2981, Ownable2Step, Reentrancy
         NUM_TOKENS = globalWalletLimit.length;
         _maxMintableSupply = maxMintableSupply;
         _globalWalletLimit = globalWalletLimit;
-        MINT_CURRENCY = mintCurrency;
+        _cosigner = cosigner;
+        _timestampExpirySeconds = timestampExpirySeconds;
         _transferable = true;
+
+        MINT_CURRENCY = mintCurrency;
         FUND_RECEIVER = fundReceiver;
 
         _setDefaultRoyalty(royaltyReceiver, royaltyFeeNumerator);
@@ -128,6 +141,21 @@ contract ERC1155M is IERC1155M, ERC1155Supply, ERC2981, Ownable2Step, Reentrancy
      */
     function removeAuthorizedMinter(address minter) external onlyOwner {
         _authorizedMinters[minter] = false;
+    }
+
+    /**
+     * @dev Returns cosign nonce.
+     */
+    function getCosignNonce(address minter, uint256 tokenId) public view returns (uint256) {
+        return totalMintedByAddress(minter)[tokenId];
+    }
+
+    /**
+     * @dev Sets cosigner.
+     */
+    function setCosigner(address cosigner) external onlyOwner {
+        _cosigner = cosigner;
+        emit SetCosigner(cosigner);
     }
 
     /**
@@ -325,14 +353,18 @@ contract ERC1155M is IERC1155M, ERC1155Supply, ERC2981, Ownable2Step, Reentrancy
      *
      * tokenId - token id 
      * qty - number of tokens to mint
-     * proof - the merkle proof generated on client side. This applies if using whitelist.
+     * proof - the merkle proof generated on client side. This applies if using whitelist
+     * timestamp - the current timestamp
+     * signature - the signature from cosigner if using cosigner
      */
     function mint(
         uint256 tokenId,
         uint32 qty,
-        bytes32[] calldata proof
+        bytes32[] calldata proof,
+        uint64 timestamp,
+        bytes calldata signature
     ) external payable virtual nonReentrant {
-        _mintInternal(msg.sender, tokenId, qty, 0, proof);
+        _mintInternal(msg.sender, tokenId, qty, 0, proof, timestamp, signature);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


contracts/ERC1155M.sol
```diff
@@ -523,6 +568,52 @@ contract ERC1155M is IERC1155M, ERC1155Supply, ERC2981, Ownable2Step, Reentrancy
...
+            MessageHashUtils.toEthSignedMessageHash(
+                keccak256(
+                    abi.encodePacked(
+                        address(this),
+                        minter,
+                        qty,
+                        _cosigner,
+                        timestamp,
+                        block.chainid,
+                        getCosignNonce(minter, tokenId)
```
tenthirtyone, line 592: I think there may be a potential transaction ordering issue here. @channing-magiceden Is there any kind of ordering protection here?

Would the offchain signature need to calculate all previous mints, including unredeemed/pending transactions, to have the correct nonce value? 
Can a mint from `authorizedMint` with `to` set to a user address break the nonce in an existing signature? E.g. a race condition?
What happens if a user abandons a tx that has been cosigned?
Do all successive mints depend on transaction ordering? So if a tx with a low fee does not mine, do all successive txs fail because the nonce is incorrect?

contracts/ERC1155M.sol
```diff
@@ -130,6 +143,21 @@ contract ERC1155M is IERC1155M, ERC1155Supply, ERC2981, Ownable2Step, Reentrancy
         _authorizedMinters[minter] = false;
     }
 
+    /**
+     * @dev Returns cosign nonce.
+     */
+    function getCosignNonce(address minter, uint256 tokenId) public view returns (uint256) {
+        return totalMintedByAddress(minter)[tokenId];
```
tenthirtyone, line 150: The implementation for `totalMintedByAddress` relies on a loop. Could this value be tracked as a uint that gets incremented on mint instead? It looks like the loop is capped by a constant value, is that to prevent running out of gas? Then does that mean an address can only mint so many token ids and the user must change to a new account?

contracts/ERC1155M.sol
```diff
@@ -368,7 +404,7 @@ contract ERC1155M is IERC1155M, ERC1155Supply, ERC2981, Ownable2Step, Reentrancy
         uint32 limit,
         bytes32[] calldata proof
     ) external payable onlyAuthorizedMinter {
-        _mintInternal(to, tokenId, qty, limit, proof);
+        _mintInternal(to, tokenId, qty, limit, proof, 0, bytes("0"));
```
tenthirtyone, line 407: How will `assertValidCosign` pass for an `authorizedMinter` if the timestamp/signature is always empty? Would this mean the `authorizedMint` function will always fail if a `_cosigner` is set?

contracts/ERC1155M.sol
```diff
@@ -583,6 +674,14 @@ contract ERC1155M is IERC1155M, ERC1155Supply, ERC2981, Ownable2Step, Reentrancy
         if (start >= end) revert InvalidStartAndEndTimestamp();
     }
 
+    /**
+     * @dev Validates the timestamp is not expired.
+     */
+    function _assertValidTimestamp(uint64 timestamp) internal view {
+        if (timestamp < block.timestamp - _timestampExpirySeconds)
```
tenthirtyone, line 681: I don't think block producers gaming the timestamp is a particular risk here but since we're cross chain am I correct that the timestamp was fast/convenient vs using block number and having to calculate a bespoke value based on each chains block time and we accepted the timestamp risk since that's low?

contracts/ERC1155M.sol
```diff
@@ -379,9 +415,18 @@ contract ERC1155M is IERC1155M, ERC1155Supply, ERC2981, Ownable2Step, Reentrancy
...
+        bytes32[] calldata proof,
+        uint64 timestamp,
+        bytes memory signature
     ) internal hasSupply(tokenId, qty) {
         uint64 stageTimestamp = uint64(block.timestamp);
+
+        if (_cosigner != address(0)) {
+            assertValidCosign(msg.sender, tokenId, qty, timestamp, signature);
+            _assertValidTimestamp(timestamp);
+            stageTimestamp = timestamp;
```
tenthirtyone, line 427: Because of this line, wouldn't it be possible for the cosigner to rewind the minting stages? For example, the only thing validating the timestamp in `_assertValidTimestamp` is that the timestamp value in the payload is < the current block timestamp. So after one stage has passed, would this value be game-able to mint after the minting window has passed?

tenthirtyone (commented):
@channing-magiceden First pass questions/comments added.

contracts/ERC1155M.sol
```diff
@@ -523,6 +568,52 @@ contract ERC1155M is IERC1155M, ERC1155Supply, ERC2981, Ownable2Step, Reentrancy
...
+    function getCosignDigest(
+        address minter,
+        uint256 tokenId,
+        uint32 qty,
+        uint64 timestamp
+    ) public view returns (bytes32) {
+        if (_cosigner == address(0)) revert CosignerNotSet();
+
+        return
+            MessageHashUtils.toEthSignedMessageHash(
```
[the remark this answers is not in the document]
channing-magiceden: This is to keep consistent with our 721M and 721CM, which use `toEthSignedMessageHash` . This is to minimize the change on cosigner service: https://github.com/channing-magiceden/erc721m/blob/cce168e12dc9c132b0bd93ffc00ef35191fd18c3/cosign-server/src/index.ts

contracts/ERC1155M.sol
channing-magiceden: Good call.

> Would the offchain signature need to calculate all previous mints, including unredeemed/pending transactions, to have the correct nonce value?
Yes offchain signature needs to count all preivous mints.

> Can a mint from authorizedMint with to set to a user address break the nonce in an existing signature? E.g. a race condition?
Yes, minting activity will bump nonce so it invalidates the old signature. This is the point of cosigning - we want to rate limit and prevent bots by controlling the cosigning process. Our UI / API will fetch new signature in that case. e.g. https://github.com/magiceden/magiceden/blob/84a82917a97cd6924a0c2fcbeb4cc42a3d73342d/js/packages/api/src/v3/routes/frame.route.ts#L67-L78

> What happens if a user abandons a tx that has been cosigned?
If mint fails or not happen, the nonce won't change. User can reuse the signature (if not expired), or grab new signature from cosign service.

> Do all successive mints depend on transaction ordering?
I don't think so. This nonce is not bond to tx but to `totalMintedByAddress`. If previous tx fails, the nonce won't bump.

contracts/ERC1155M.sol
channing-magiceden: Yeah good call. I think we can use single mapping `mapping(address => uint32)` (total tokens minted per wallet). This may save some gas (pending experiment). I can test it in following up PR.

The cap is `NUM_TOKENS`. It's the number of "editions" of a ERC1155 collection. It's fixed / determined in constructor. Each edition can has unlimited copies.

contracts/ERC1155M.sol
channing-magiceden: `authorizedMint` will still work as Crossmint and Reservoir Relay allow us to pass arbitrary data, in this case, timestamp and signature.

contracts/ERC1155M.sol
channing-magiceden: Good call, we never thought about cross chain scenario. The timestamp check is just to ensure freshness of signature. The good thing is we don't have crosschain use case yet (we were having layerzero cross mint but we turned it down). We should revisit if we want to do cross mint again.

contracts/ERC1155M.sol
channing-magiceden: Yeah good call. This behavior is intentional. We want to use payload time (guarded by signature) as "mint time". Previously, we have seen many users failed to mint. When they grab signature, the stage is still valid, but when they submit tx, stage is ended. The user experience for them is very bad so we decided to use they timestamp when they grab signature. It's more like extend extra few minutes for those who intent to mint. We still have other guards such as whitelist, supply etc.

tenthirtyone (approved):

contracts/ERC721CM.sol
```diff
@@ -588,11 +594,12 @@ contract ERC721CM is IERC721M, ERC721ACQueryable, Ownable, ReentrancyGuard {
     }
 
     /**
-     * @dev Returns data hash for the given minter, qty and timestamp.
+     * @dev Returns data hash for the given minter, qty, waiveMintFee and timestamp.
      */
     function getCosignDigest(
         address minter,
         uint32 qty,
+        bool waiveMintFee,
```
nothing0012, line 602: do we have ongoing mints that will break the cosigner server?

nothing0012 (approved):

contracts/ERC721CM.sol
channing-magiceden: `getCosignDigest` is only invoked inside contract - we could have mark it internal actually. `getCosignNonce` is the only method called by our FE and it keeps the unchanged for 721 / newly added for 1155. 

Re cosigning service, the interface changed and `waiveMintFee` is optional:
```
export interface ICosignRequest {
  collectionContract: string;
  minter: string;
  qty: number;
  chainId: number;
  nonce: number;
  waiveMintFee?: boolean;
}
```

So it should be safe for ongoing mints.

cosign-server/src/types.ts
```diff
@@ -4,6 +4,7 @@ export interface ICosignRequest {
   qty: number;
   chainId: number;
   nonce: number;
+  waiveMintFee?: boolean;
```
channing-magiceden, line 7: this is optional in cosign service request payload

nothing0012 (approved):

PR #152, opened 2024-09-19
magicdrop refactor and contract factory

wolfy-nft:
- setup foundry
- add contract factory
- various refactors for code quality and file structure

hardhat tests passing
Screenshot 2024-09-19 at 12 59 28 PM


test/ERC721CM.test.ts
```diff
@@ -1991,14 +1991,4 @@ describe('ERC721CM', function () {
       );
     });
   });
-
-  describe('Transfer validator', function () {
-    it('default validator settings', async () => {
-      expect(await contract.getTransferValidator()).to.equal(
-        '0x0000000000000000000000000000000000000000',
```
wolfy-nft, line 1998: this is set automatically in the newest version

contracts/common/interfaces/IInitializableToken.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.19;
+
+interface IInitializableToken {
```
wolfy-nft, line 4: todo: remove this

contracts/nft/MagicDropERC1155Initializable.sol
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+import {ERC1155Upgradeable} from "@openzeppelin/contracts-upgradeable/token/ERC1155/ERC1155Upgradeable.sol";
+import {OwnableUpgradeable} from "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+import {IInitializableToken} from "../common/interfaces/IInitializableToken.sol";
+
+contract MagicDropERC1155Initializable is ERC1155Upgradeable, OwnableUpgradeable, IInitializableToken {
```
wolfy-nft, line 8: delete this

contracts/nft/MagicDropERC721Initializable.sol
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+import {ERC721Upgradeable} from "@openzeppelin/contracts-upgradeable/token/ERC721/ERC721Upgradeable.sol";
+import {OwnableUpgradeable} from "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+import {IInitializableToken} from "../common/interfaces/IInitializableToken.sol";
+
+contract MagicDropERC721Initializable is ERC721Upgradeable, OwnableUpgradeable, IInitializableToken {
```
wolfy-nft, line 8: delete this

contracts/nft/erc1155m/ERC1155MStorage.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+import {MintStageInfo1155} from "../../common/Structs.sol";
+
+contract ERC1155MStorage {
+    uint256 public constant VERSION = 1;
```
wolfy-nft, line 7: setup better versioning outside of this storage contract

contracts/common/AuthorizedMinterControl.sol
```diff
@@ -0,0 +1,61 @@
+pragma solidity ^0.8.20;
+
+/**
+ * @title AuthorizedMinterControl
+ * @dev Abstract contract to manage authorized minters for ERC1155M tokens
+ */
+abstract contract AuthorizedMinterControl {
```
wolfy-nft, line 7: i think we should eventually add Ownable to this, but we have mixed versions of Ownable in the codebase so it will break things. I've opted to make the addAuthorizedMinter functions virtual so the implementing contract can add its own access control

scripts-foundry/common/DeployMagicDropImplementation.s.sol
```diff
@@ -0,0 +1,63 @@
...
+        TokenStandard standard = parseTokenStandard(vm.envString("TOKEN_STANDARD"));
+        string memory version = vm.envString("CONTRACT_VERSION");
+        uint256 privateKey = vm.envUint("PRIVATE_KEY");
+
+        address deployedAddress;
+
+        console.log("Version: %s", version);
+        console.log("Expected Address: %s", expectedAddress);
+        console.log("Salt: %s", uint256(salt));
+        console.log("hello");
```
wolfy-nft, line 26: cleanup

contracts/factory/MagicDropCloneFactory.sol
```diff
@@ -0,0 +1,256 @@
...
+            mstore(0x20, MAGICDROP_FACTORY_STORAGE)
+            sstore(keccak256(0x00, 0x40), 1) // usedSalts[salt] = true
+        }
+
+        emit NewContractInitialized({
+            contractAddress: instance,
+            initialOwner: initialOwner,
+            implId: implId,
+            standard: standard,
+            name: name,
```
raymond-me, line 149: Screenshot 2024-10-08 at 9 50 55 PM


I get a stack too deep error when running `forge build` 

bypassed by adding `via_ir = true` to `foundry.toml`

wolfy-nft: yeah im aware of this, i can add via_ir to the foundry, though i was hoping future refactors can fix this issue all together. i think this must be set in hardhat already

raymond-me (commented):
First pass didn't find anything major issues with the contract updates. Haven't review deployment scripts, but will do that in another pass.

contracts/nft/erc1155m/ERC1155MInitializableV1_0_0.sol
```diff
@@ -0,0 +1,583 @@
...
+            stageTimestamp = timestamp;
+        }
+
+        uint256 activeStage = getActiveStageFromTimestamp(stageTimestamp);
+
+        MintStageInfo1155 memory stage = _mintStages[activeStage];
+        uint80 adjustedMintFee = waiveMintFee ? 0 : stage.mintFee[tokenId];
+
+        // Check value if minting with ETH
+        if (_mintCurrency == address(0) && msg.value < (stage.price[tokenId] + adjustedMintFee) * qty) {
```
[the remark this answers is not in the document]
wolfy-nft: yeah lets mark this as an improvement for the next refactor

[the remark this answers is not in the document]
wolfy-nft: technically shouldnt be possible unless they mint from the contract directly

remappings.txt
```diff
@@ -0,0 +1,8 @@
+solady/=lib/solady/
+solemate/=/lib/solemate/src/
+@openzeppelin/contracts/=lib/openzeppelin-contracts/contracts/
+@openzeppelin/contracts-upgradeable/=lib/openzeppelin-contracts-upgradeable/contracts/
+forge-std/=lib/forge-std/src/
+erc721a/contracts=lib/ERC721a/contracts/
```
tenthirtyone, line 6: This will work on mac, on linux the lowercase "a" in `lib/ERC721a/contracts` will not map properly because of case sensitivity. 

Probably not a problem in local development on a mac but in a CICD env running an ext4 filesystem this will throw a file not found.

Suggested: `lib/ERC721a/contracts`

contracts/nft/erc1155m/ERC1155MInitializableV1_0_0.sol
```diff
@@ -0,0 +1,580 @@
...
+
+import {MintStageInfo1155} from "../../common/Structs.sol";
+import {MINT_FEE_RECEIVER} from "../../utils/Constants.sol";
+import {IERC1155M} from "./interfaces/IERC1155M.sol";
+import {ERC1155MStorage} from "./ERC1155MStorage.sol";
+import {Cosignable} from "../../common/Cosignable.sol";
+import {AuthorizedMinterControl} from "../../common/AuthorizedMinterControl.sol";
+
+/// @title ERC1155MInitializableV1_0_0
+/// @notice An upgradeable ERC1155 contract with multi-stage minting, royalties, and authorized minters
```
raymond-me, line 22: Suggest:
`@notice An upgradeable ERC1155 contract `
`@notice An initializable ERC1155 contract`

Unless this implementation contract is intended to be fully upgradeable.

contracts/nft/erc721m/ERC721M.sol
```diff
@@ -0,0 +1,465 @@
...
+import {ERC721AQueryable, IERC721A, ERC721A} from "erc721a/contracts/extensions/ERC721AQueryable.sol";
+import {IERC721M} from "./interfaces/IERC721M.sol";
+import {ERC721MStorage} from "./ERC721MStorage.sol";
+import {Cosignable} from "../../common/Cosignable.sol";
+import {AuthorizedMinterControl} from "../../common/AuthorizedMinterControl.sol";
+import {MintStageInfo} from "../../common/Structs.sol";
+import {MINT_FEE_RECEIVER} from "../../utils/Constants.sol";
+
+/// @title ERC721M
+/// @notice An upgradeable ERC721 contract with multi-stage minting, royalties, and authorized minters
```
raymond-me, line 19: Suggest updating wording to from upgradeable to initializable

contracts/factory/MagicDropCloneFactory.sol
```diff
@@ -0,0 +1,245 @@
...
+
+    /// @notice Creates a new deterministic clone of a MagicDrop contract
+    /// @param name The name of the new contract
+    /// @param symbol The symbol of the new contract
+    /// @param standard The token standard of the new contract
+    /// @param initialOwner The initial owner of the new contract
+    /// @param implId The implementation ID
+    /// @param salt A unique salt for deterministic address generation
+    /// @return The address of the newly created contract
+    function createContractDeterministic(
```
tomhirst, line 87: This might become obvious as I read more of the code but why do we need to know the contract's address?

contracts/mocks/TestStaking.sol
```diff
@@ -1,22 +1,20 @@
...
 
     constructor(address nft) {
         _nft = IERC721A(nft);
     }
 
-    function isStaked(
-        address staker,
-        uint256 tokenId
-    ) public view returns (bool) {
+    function isStaked(address staker, uint256 tokenId) public view returns (bool) {
```
tomhirst, line 17: Is there a use case for wanting to know if a token is staked without having to pass in the owner's address?

contracts/nft/creator-token-standards/ERC721ACQueryable.sol
```diff
@@ -0,0 +1,52 @@
...
+                ++i;
+            }
+        }
+    }
+
+    function _msgSenderERC721A() internal view virtual override returns (address) {
+        return _msgSender();
+    }
+
+    function _tokenType() internal pure override returns (uint16) {
```
tomhirst, line 49: Where is this used?

contracts/nft/erc721m/extensions/BucketAuction.sol
```diff
@@ -30,9 +28,10 @@ contract BucketAuction is IBucketAuction, ERC721M {
         uint256 maxMintableSupply,
         uint256 globalWalletLimit,
         address cosigner,
+        uint256 timestampExpirySeconds,
         uint256 minimumContributionInWei,
-        uint64 startTimeUnixSeconds,
-        uint64 endTimeUnixSeconds,
+        uint256 startTimeUnixSeconds,
+        uint256 endTimeUnixSeconds,
```
tomhirst, line 34: Can we save gas by staying lower on these? Was `uint64` future proof enough?

contracts/nft/erc1155m/ERC1155MStorage.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+import {MintStageInfo1155} from "../../common/Structs.sol";
+
+contract ERC1155MStorage {
```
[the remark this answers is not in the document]
wolfy-nft: we could but its not necessary yet since it has no functions to override, im planning another refactor of the storage contract to remove some of the duplicate functions in the 721/1155 contracts

contracts/mocks/TestStaking.sol
wolfy-nft: idk the context of this test, its an old one that was formatted by forge

contracts/nft/creator-token-standards/ERC721ACQueryable.sol
wolfy-nft: its a requirement of CreatorTokenBase to override this function

contracts/nft/erc721m/extensions/BucketAuction.sol
wolfy-nft: yeah we had uint64 originally, but i purposely converted them to 256 because they are being compared against block.timestamp which is uint256. we were having to convert them anyway

tenthirtyone (commented):
First pass

tenthirtyone:
Inheritance graph looking knee-ice

inheritance

tenthirtyone:
piggybacking on @raymond-me 's `via-ir` -> This will affect static analysis as well. So some tooling (slither security tool) may have issues with `via-ir`

tenthirtyone:
Possibly unrelated, because it does not seem to be part of the changes, but when I run hardhat tests I get an "unable to estimate gas" error on this line

https://github.com/me-foundation/magicdrop/blob/757d3c502172fdba7139ec3490dc85ede07b28bf/test/erc1155m/ERC1155M.test.ts#L70

When I override the `gasLimit` to something like 3_000_000 I get a "reverted with custom error `Unauthorized`" coming from `Ownable` for `setDefaultRoyalty`

```
    Error: VM Exception while processing transaction: reverted with custom error 'Unauthorized()'
    at ERC1155M._checkOwner (solady/src/auth/Ownable.sol:162)
    at ERC1155M.onlyOwner (solady/src/auth/Ownable.sol:279)
    at ERC1155M.setDefaultRoyalty (contracts/nft/erc1155m/ERC1155M.sol:284)
    at ERC1155M.constructor (contracts/nft/erc1155m/ERC1155M.sol:72)
```

scripts-foundry/common/DeployMagicDropTokenImplRegistry.s.sol
```diff
@@ -0,0 +1,30 @@
...
+
+        // Deploy the contract using CREATE2 directly
+        MagicDropTokenImplRegistry registryImpl = new MagicDropTokenImplRegistry{salt: salt}();
+
+        // Verify the deployed address matches the predicted address
+        if (address(registryImpl) != expectedAddress) {
+            revert AddressMismatch();
+        }
+
+        registryImpl.initialize(initialOwner);
```
tenthirtyone, line 26: `initialize` should be coupled with the deployment transaction. 

Initialize is wide open and can only be called once. I'm used to seeing the initialization params encoded and passed to the deployment function so that after deployment the initialize function is called in the same transaction. 

This prevents outside actors from calling initialize in between the transactions

scripts-foundry/common/DeployMagicDropCloneFactory.s.sol
```diff
@@ -0,0 +1,30 @@
...
+        vm.startBroadcast(privateKey);
+        // Deploy the contract using CREATE2 directly
+        MagicDropCloneFactory factoryImpl = new MagicDropCloneFactory{salt: salt}();
+
+        // Verify the deployed address matches the predicted address
+        if (address(factoryImpl) != expectedAddress) {
+            revert AddressMismatch();
+        }
+
+        factoryImpl.initialize(initialOwner, registry);
```
tenthirtyone, line 26: https://github.com/me-foundation/magicdrop/pull/152/files#r1799916516

contracts/factory/MagicDropCloneFactory.sol
```diff
@@ -0,0 +1,245 @@
...
+import {Initializable} from "solady/src/utils/Initializable.sol";
+import {Ownable} from "solady/src/auth/Ownable.sol";
+import {LibClone} from "solady/src/utils/LibClone.sol";
+import {TokenStandard} from "../common/Structs.sol";
+import {MagicDropTokenImplRegistry} from "../registry/MagicDropTokenImplRegistry.sol";
+
+/// @title MagicDropCloneFactory
+/// @notice A factory contract for creating and managing clones of MagicDrop contracts
+/// @dev This contract uses the UUPS proxy pattern and is initializable
+contract MagicDropCloneFactory is Initializable, Ownable, UUPSUpgradeable {
```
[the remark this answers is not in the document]
wolfy-nft: so do you recommend adding receive functions?

contracts/nft/erc1155m/ERC1155MStorage.sol
```diff
@@ -0,0 +1,22 @@
...
+    // Mint stage information. See MintStageInfo for details.
+    MintStageInfo1155[] internal _mintStages;
+
+    string public name;
+    string public symbol;
+    bool internal _transferable; // Whether the token can be transferred.
+    uint256[] internal _maxMintableSupply; // The total mintable supply per token.
+    uint256[] internal _globalWalletLimit; // Global wallet limit, across all stages, per token.
+    uint256 internal _totalMintFee;
+    address internal _mintCurrency; // If 0 address, use native currency.
```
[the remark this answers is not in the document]
wolfy-nft: we have setters for most of these, if we make them immutable, ops wont be able to update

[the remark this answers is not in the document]
wolfy-nft: my next refactor plan involves moving setters into this storage contract do redupe those functions across the repo

contracts/nft/erc721m/ERC721MStorage.sol
```diff
@@ -0,0 +1,45 @@
...
+pragma solidity ^0.8.20;
+
+import "../../common/Structs.sol";
+
+contract ERC721MStorage {
+    // Whether this contract is mintable.
+    bool internal _mintable;
+
+    // Whether base URI is permanent. Once set, base URI is immutable.
+    bool internal _baseURIPermanent;
```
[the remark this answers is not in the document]
wolfy-nft: ah ya i removed this, just have missed it in some spots

scripts-foundry/common/DeployMagicDropTokenImplRegistry.s.sol
wolfy-nft: any examples you can share?

I believe with foundry these two calls are done in one transaction

tenthirtyone: Ah I took for granted how Foundry does deploy scripts.

So it looks like the transactions are dependent on Foundry's CREATE2 deployer (0x4e59b44847b379578588920cA78FbF26c0B4956C) so that would make sense that the data could be encoded/called as part of the init bytecode. 

It would be important to note that if we are targeting uniform contract addresses across chains then we are dependent on their CREATE2 contract. I think this is a trade off worth thinking about. 

Here are some `upgradeToAndCall` examples:
[Reference Implementation](https://eips.ethereum.org/EIPS/eip-1967)
[Aave](https://github.com/search?q=repo%3Aaave%2Faave-v3-core%20upgradeToAndCall&type=code)
[Open Zeppelin](https://github.com/search?q=repo%3AOpenZeppelin%2Fopenzeppelin-contracts%20upgradeToAndCall&type=code)

I usually lean on OZ unit tests for implementation details.

contracts/factory/MagicDropCloneFactory.sol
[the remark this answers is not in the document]
tenthirtyone: nah that would just get more eth in the contracts. I think this can be fixed bumping open zeppelin contracts. But I haven't looked over all the changes. The current version is for sol 0.8.4. Latest OZ is 0.8.20 and doesn't have the payable functions. 

I guess there are a couple options -> Bumping OZ will effect the entire repo. Overriding all the functions with unintended behavior would not be fun. Accepting the risk of locked eth kind of leaves the problem as-is. And adding a withdrawal may be benign/never be used and is kind of just a patch. Interested in your thoughts, maybe I missed an option.

contracts/nft/erc1155m/ERC1155MStorage.sol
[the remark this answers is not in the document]
tenthirtyone: I guess I am missing the setters. I can only see these variables being used in the constructors. can you drop me a permalink?

contracts/factory/MagicDropCloneFactory.sol
```diff
@@ -0,0 +1,245 @@
...
+
+            if saltUsed {
+                // revert SaltAlreadyUsed()
+                mstore(0x00, 0x0ced3043)
+                revert(0x1c, 0x04)
+            }
+        }
+
+        // Predict the address where the contract will be deployed
+        address predictedAddress = LibClone.predictDeterministicAddress(impl, salt, address(this));
```
tenthirtyone, line 117: It may be cheaper to do this off chain. There's a thread somewhere about gas efficiency of making comparisons vs just letting the tx fail. It would be interesting to see the gas usage of each.

contracts/nft/erc1155m/ERC1155MStorage.sol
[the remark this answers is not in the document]
wolfy-nft: Oh you are right, I thought these had setters. Wdyt about adding them? At least for updating fund receiver, and mint currency

contracts/factory/MagicDropCloneFactory.sol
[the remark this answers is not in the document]
wolfy-nft: I'm ok with the risk of locked funds

contracts/factory/MagicDropCloneFactory.sol
```diff
@@ -0,0 +1,256 @@
...
+    ==============================================================*/
+
+    MagicDropTokenImplRegistry private _registry;
+    bytes4 private constant INITIALIZE_SELECTOR = bytes4(keccak256("initialize(string,string,address)"));
+
+    /*==============================================================
+    =                            STRUCTS                           =
+    ==============================================================*/
+
+    struct MagicDropFactoryStorage {
```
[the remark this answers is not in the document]
wolfy-nft: yup i will remove these extra checks

scripts-foundry/common/DeployMagicDropTokenImplRegistry.s.sol
tenthirtyone: I think from our convo offline relying on Foundry should be fine. If we ever run into issues with the CREATE2 dependency we can major version bump and pick the convo back up from that point -> e.g. maintaining our own factory deployer. I think the effort v return right now is low.

contracts/nft/erc1155m/ERC1155MStorage.sol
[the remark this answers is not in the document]
tenthirtyone: Yeah, I agree with you.

cli/cmds/cleaned-whitelist.json
channing-magiceden, line 1: Is this a temp file?

cli/cmds/utils
```diff
@@ -0,0 +1,589 @@
...
+        echo "Your local main branch is ahead of the remote."
+    else
+        echo "Your local main branch has diverged from the remote."
+        exit 1
+    fi
+
+    forge build --root "$BASE_DIR"
+}
+
+# TODO(adam) finish deployment loading
```
channing-magiceden, line 497: Same here.

cli/cmds/utils
```diff
@@ -0,0 +1,589 @@
...
+        42161) RPC_URL="https://arb1.arbitrum.io/rpc" ;; # Arbitrum
+        1329) RPC_URL="https://evm-rpc.sei-apis.com" ;; # Sei
+        33139) RPC_URL="https://curtis.rpc.caldera.xyz/http" ;; # ApeChain
+        *) echo "Unsupported chain id"; exit 1 ;;
+    esac
+
+    export RPC_URL
+}
+
+generate_merkle_root() {
```
channing-magiceden, line 331: TBH I hesitate to agree bash script is the best option in this case. This example demonstrates that the complexity of our scripts/cmds is too high for bash to handle elegantly:

- increased complexity - the merkle root logic lives in `generateMerkleRoot.ts` 
- hard to build / debug:
  - `generateMerkleRootPath` is hardcoded. If file is moved, the issue can only be caught in runtime.
  - result need to be filtered using `sed` and regex
- readability is not high:
  - e.g. it's impossible to know the meaning of `false` as the second param to `generateMerkleRootPath`

cli/cmds/utils
```diff
@@ -0,0 +1,589 @@
...
+
+
+    if [[ $? -ne 0 ]]; then
+        echo "Error: Failed to parse stages JSON"
+        return 1
+    fi
+
+    stages_data=$(echo "$stages_data" | sed 's/"//g') # remove any quotes
+
+    # Replace inner square brackets with parentheses, but keep the outer brackets
```
channing-magiceden, line 393: This is another example. It's very hard to understand and modify this transforming logic.

nothing0012:
can you help fix the CI?

image

PR #156, opened 2024-10-02
Unify mint functions

channing-magiceden:
- Merge `mintWithLimit` into `mint`
- Merge `crossmint` into `authorizedMint`

`ERC721CMInitializable` is not modified as it needs more updates.

### DIFF
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -80,6 +80,11 @@ const config: HardhatUserConfig = {
         accountsBalance: '1000000000000000000000'
       }
     },
+    apechain: {
+      url: 'https://apechain.calderachain.xyz/http',
+      accounts:
+        process.env.PRIVATE_KEY !== undefined ? [process.env.PRIVATE_KEY] : [],
+    },
     base: {
       url: process.env.BASE_URL || '',
       accounts:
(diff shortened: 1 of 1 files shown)


contracts/ERC721CM.sol
```diff
@@ -336,29 +328,12 @@ contract ERC721CM is IERC721M, ERC721ACQueryable, Ownable, ReentrancyGuard {
      * @dev Mints token(s).
      *
      * qty - number of tokens to mint
-     * proof - the merkle proof generated on client side. This applies if using whitelist.
-     * timestamp - the current timestamp
-     * signature - the signature from cosigner if using cosigner.
-     */
-    function mint(
```
nothing0012, line 343: for existing mints, how does the client know which `mint` function to use? given that the function signature changed.

test/ERC721CMRoyaltiesClone.test.ts
```diff
@@ -401,7 +401,7 @@ describe('ERC721CMIRoyaltiesClone', function () {
       await expect(mint).to.be.revertedWith('NotEnoughValue');
     });
 
-    it('revert on reentrancy', async () => {
+    it.skip('revert on reentrancy', async () => {
```
XTHuo-MagicEden, line 404: Would we fix the test here later?

PR #168, opened 2025-01-23
ERC1155 ZKSync

tenthirtyone:
Apologies, the solidity formatter hit some files. 

Changes:
- Update .gitignore with coverage and zkout dir
- replaces
`import {ERC1155} from "solady/src/tokens/ERC1155.sol";`
with
`import {ERC1155} from "solady/src/tokens/ext/zksync/ERC1155.sol";`
- Bumps solady to 0.0.298 in foundry and npm


nothing0012 (approved):

nothing0012:
good to know that solady has zksync ext.

does it work for non-zksync chains?

tenthirtyone:
@nothing0012 I think it depends. This contract should, but code related to zksync-specific features (system contracts, account abstraction, bridging) won't and some EVM opcodes are inefficient on zksync and not implemented.

PR #171, opened 2025-01-27
update magicdrop cli for Abstract

wolfy-nft:


nothing0012 (approved):
I feel like the PR can be more generic to support zksync chains, but can be done in the future PRs

nothing0012 (approved):

PR #173, opened 2025-02-05
Add mintFee functionality

tomhirst:
**Global mint fee functionality**

- Adds `implementationId => mintFee` mapping in `RegistryData` struct to manage `mintFees` on an implementation level
- When creating a clone in `MagicDropCloneFactory`, get `mintFee` for the `implId` from registry storage and pass to the cloned contract's initializer
- Sets immutable `mintFee` on cloned contract
- Updates `RequiredValue` checks on mint 
- Updates logic in `_splitProceeds` to ensure `mintFee` is distributed to `MINT_FEE_RECEIVER`
- Adds various getters, setters, utils and updates to support `mintFee` functionality
- Refactors launchpad contracts to use global `mintFee` over `stage.mintFee`
- Bumps launchpad contract versions to `V1_0_2`
- Removes % based `protocolFee` in self-serve contracts
- Updates tests

**General improvements**

- added Initializable to MagicDropCloneFactory and MagicDropTokenTokenImplRegistry
- converted constructor to initializer
- disable initializers from constructor
- added storage gap __gap to MagicDropCloneFactory and MagicDropTokenTokenImplRegistry
- MagicDropCloneFactory  updated createContractDeterministic to re-hash the public salt with the senders address and chain id
- Replaced Ownable with OwnableRoles in MagicDropTokenTokenImplRegistry
- added a manager role

Screenshot 2025-02-12 at 14 27 53

Screenshot 2025-02-12 at 18 25 25


contracts/nft/erc1155m/clones/ERC1155MagicDropCloneable.sol
```diff
@@ -420,16 +435,28 @@ contract ERC1155MagicDropCloneable is ERC1155MagicDropMetadataCloneable {
             revert PayoutRecipientCannotBeZeroAddress();
         }
 
+        uint256 proceeds = msg.value;
+
+        if (mintFee > 0) {
+            proceeds -= mintFee;
```
XTHuo-MagicEden, line 441: Should we consider the `qty` here? If I mint 2, we should charge mintFee * 2?

tomhirst: Yep. Amended.

contracts/factory/MagicDropCloneFactory.sol
```diff
@@ -146,11 +157,15 @@ contract MagicDropCloneFactory is Ownable, UUPSUpgradeable {
...
+        // Retrieve the mint fee for the implementation
+        uint256 mintFee = _registry.getMintFee(standard, implId);
+
         // Create a non-deterministic clone of the implementation contract
         address instance = LibClone.clone(impl);
 
         // Initialize the newly created contract
-        (bool success,) = instance.call(abi.encodeWithSelector(INITIALIZE_SELECTOR, name, symbol, initialOwner));
+        (bool success,) =
+            instance.call(abi.encodeWithSelector(INITIALIZE_SELECTOR, name, symbol, initialOwner, mintFee));
```
wolfy-nft, line 168: you need to add mintFee to the INITIALIZE_SELECTOR at the top of the file or this call will fail

wolfy-nft:
```
 bytes4 private constant INITIALIZE_SELECTOR = bytes4(keccak256("initialize(string,string,address,uint256)"));
 ```

File: contracts/nft/erc1155m/clones/ERC1155MagicDropCloneable.sol [hunk 3]
```diff
@@ -87,6 +87,14 @@ contract ERC1155MagicDropCloneable is ERC1155MagicDropMetadataCloneable {
     /// @notice Only addresses proven by a valid Merkle proof can mint during this stage.
     mapping(uint256 => AllowlistStage) internal _allowlistStages; // tokenId => allowlistStage
 
+    /// @dev The address that receives mint fees.
+    /// @notice This is fixed and cannot be changed.
+    address public constant MINT_FEE_RECIPIENT = 0xA3833016a4eC61f5c253D71c77522cC8A1cC1106;
```
wolfy-nft, line 92: we already have a `MINT_FEE_RECEIVER` in `contracts/utils/Constants.sol` used by our launchpad contracts. if we cant re-use it, move this to the constants file and call something like `SELF_SERVE_MINT_FEE_RECEIVER`

contracts/nft/erc1155m/clones/ERC1155MagicDropCloneable.sol
```diff
@@ -415,21 +431,33 @@ contract ERC1155MagicDropCloneable is ERC1155MagicDropMetadataCloneable {
...
+            proceeds -= (mintFee * qty);
+            SafeTransferLib.safeTransferETH(MINT_FEE_RECIPIENT, mintFee);
+        }
+
+        // If there are no remaining proceeds after mint fee is taken, exit early
+        if (proceeds == 0) {
+            return;
+        }
+
         if (PROTOCOL_FEE_BPS > 0) {
```
wolfy-nft, line 451: we should replace `PROTOCOL_FEE_BPS` with mintFee

contracts/nft/erc721m/clones/ERC721MagicDropCloneable.sol
```diff
@@ -91,6 +91,14 @@ contract ERC721MagicDropCloneable is ERC721MagicDropMetadataCloneable {
     /// @notice Only addresses proven by a valid Merkle proof can mint during this stage.
     AllowlistStage private _allowlistStage;
 
+    /// @dev The address that receives mint fees.
+    /// @notice This is fixed and cannot be changed.
+    address public constant MINT_FEE_RECIPIENT = 0xA3833016a4eC61f5c253D71c77522cC8A1cC1106;
```
wolfy-nft, line 96: another case to move into constants, so we have one source of truth

contracts/registry/MagicDropTokenImplRegistry.sol
```diff
@@ -216,8 +238,10 @@ contract MagicDropTokenImplRegistry is UUPSUpgradeable, Ownable, IMagicDropToken
         $.tokenStandardData[standard].implementations[implId] = impl;
         $.tokenStandardData[standard].nextImplId = implId + 1;
         $.tokenStandardData[standard].deploymentFees[implId] = deploymentFee;
+        $.tokenStandardData[standard].mintFees[implId] = mintFee;
         emit ImplementationRegistered(standard, impl, implId, deploymentFee);
         emit DeploymentFeeSet(standard, implId, deploymentFee);
+        emit MintFeeSet(standard, implId, mintFee);
```
wolfy-nft, line 244: come to think of it, maybe we should put everything in `ImplementationRegistered` event and delete DeploymentFeeSet and MintFeeSet

contracts/factory/MagicDropCloneFactory.sol
tomhirst: Good catch.

contracts/nft/erc1155m/clones/ERC1155MagicDropCloneable.sol
tomhirst: Isn't it better to keep both options? A flat mint fee is different to a % cut of the proceeds. There may be instances where we want 1 of, or both, functionalities.

contracts/nft/erc721m/clones/ERC721MagicDropCloneable.sol
tomhirst: Can move to shared constant, looks like https://etherscan.io/address/0x0B98151bEdeE73f9Ba5F2C7b72dEa02D38Ce49Fc#multichain-portfolio has some mintFee proceeds already.

contracts/registry/MagicDropTokenImplRegistry.sol
tomhirst: We'd still want to fire `DeploymentFeeSet` separately on `setDeploymentFee` thought, right? Same for `mintFee`.

I'll add `mintFee` to `ImplementationRegistered` for now.

test/erc1155m/clones/ERC1155MagicDropCloneable.t.sol
```diff
@@ -29,6 +29,7 @@ contract ERC1155MagicDropCloneableTest is Test {
     uint256 internal allowlistEnd;
     address royaltyRecipient = address(0x8888);
     uint96 royaltyBps = 1000;
+    uint256 mintFee = 0; // 0 ether
```
wolfy-nft, line 32: we should make this non zero to ensure mint fee works

wolfy-nft: + add a test to ensure we get the expected fees

tomhirst: Added `mintFee` and fixed broken tests.

contracts/nft/erc721m/ERC721MStorageV1_0_2.sol
wolfy-nft, line 1: we need to version this

contracts/nft/erc1155m/clones/ERC1155MagicDropCloneable.sol
wolfy-nft: we could

wolfy-nft: actually since we are dropping % fee we can combine mint fee receiver + protocol fee address

test/factory/MagicDropCloneFactoryUpgradeTest.t.sol
```diff
@@ -0,0 +1,192 @@
...
+import {Ownable} from "solady/src/auth/Ownable.sol";
+import {ERC721MagicDropCloneable} from "../../contracts/nft/erc721m/clones/ERC721MagicDropCloneable.sol";
+import {TokenStandard} from "../../contracts/common/Structs.sol";
+
+contract MagicDropCloneFactoryV2 is MagicDropCloneFactory {
+    // New storage variable (example)
+    uint256 private _maxDeploymentFee;
+
+    // Reduce gap by 1
+    uint256[47] private __gap;
```
tenthirtyone, line 19: You can delete this line and tests will pass, here is why:

Inheritance storage mapping works differently.

The `_maxDeploymentFee` variable on storage layout comes after all of the variables in the inherited MagicDropCloneFactory and contract storage layout is first to last, from right to left.

Example:
``` solidity
- MagicDropCloneFactory
var1
var2
var3
- MagicDropCloneFactoryV2
var1
var2
```

The shadow redeclaration of the private `__gap` variable maps onto the original variable but the number of slots are determined when the var is originally instantiated. The EVM will not re-allocate storage slots like this. So the contract storage looks like this:

``` solidity
- MagicDropCloneFactory
slot0: _registry
slot1 - slot48: __gap
slot49: INITIALIZE_SELECTOR
- MagicDropCloneFactoryV2
slot50: _maxDeploymentFee
```

The shadow redeclaration effect is that this implementation will throw on index 48. However, slot48 is still reserved for `__gap` it is just unused now and will throw if we try to access the unused storage slot index (out of bounds error). 

You can delete this line because the `_maxDeploymentFee` variable does not impact the storage layout of `MagicDropCloneFactory`

The `__gap` storage strategy is not an inheritance storage strategy. Use `__gap` when you are upgrading proxy implementations and manually managing storage slots. 

Example:
``` solidity
contract MagicDropCloneFactory {
  address _registry;
  uint256[50] __gap;
  bytes4 constant BIGVARNAME = "BIGVAR";
}
```

V2 of this contract would look like this:

``` solidity
contract MagicDropCloneFactory {
  address _registry;
  uint256 _maxDeploymentFee;
  uint256[49] __gap; // now we decrement gap. The same storage allocations exist, but _maxDeploymentFee is in the previous index[0]
  bytes4 constant BIGVAR...
}
```

The same gap between `_registry` and `BIGVAR` still exists on the original storage mapping.

tenthirtyone (changes requested):

contracts/registry/MagicDropTokenImplRegistry.sol
```diff
@@ -173,13 +194,32 @@ contract MagicDropTokenImplRegistry is UUPSUpgradeable, Ownable, IMagicDropToken
         }
     }
 
+    /// @dev Gets the mint fee for a given token standard
```
tenthirtyone, line 197: personal/discussion nits: until we definitively know the gas savings we get for inlined assembly I think it makes sense to stick to the original Solidity ethos of "could be accessible to developers without specialized blockchain knowledge"

test/erc721m/ERC721CMInitializableTest.t.sol
```diff
@@ -30,6 +30,7 @@ contract ERC721CMInitializableTest is Test {
     uint256 public constant GLOBAL_WALLET_LIMIT = 0;
     address public clone;
     uint256 public startTime;
+    uint256 public mintFee = 10000000000000; // 0.00001 ether;
```
tenthirtyone, line 33: discussion nit: I always mess up my zeros. If you prefer, you can use 
`uint256 public mintFee = 0.00001 ether;`

It's saved me on precision more than once. Prior to 0.7 you could have done `1 szabo` but they removed finney's and szabos

contracts/factory/MagicDropCloneFactory.sol
```diff
@@ -93,11 +115,17 @@ contract MagicDropCloneFactory is Ownable, UUPSUpgradeable {
             revert InsufficientDeploymentFee();
         }
 
+        // Retrieve the mint fee for the implementation
+        uint256 mintFee = _registry.getMintFee(standard, implId);
+
+        // Create a unique salt by combining original salt with chain ID and sender
+        bytes32 _salt = keccak256(abi.encode(salt, block.chainid, msg.sender));
```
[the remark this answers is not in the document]
tomhirst: I don't need think it's a requirement to create clones with the same address on multiple chains, so this should be fine.

Question is: Why do we clone in a deterministic fashion in the first place? 

@wolfy-nft To confirm?

contracts/registry/MagicDropTokenImplRegistry.sol
tomhirst: I'm generally in agreement that inline assembly should be used sparsely or inherited from battle-tested libraries unless the gas savings are significant.

In this `getMintFee` case, I use it for uniformity with `getDeploymentFee`. Both getters should follow the same retrieval pattern.

contracts/registry/MagicDropTokenImplRegistry.sol
tomhirst: @wolfy-nft Can I close this?

test/factory/MagicDropCloneFactoryUpgradeTest.t.sol
tomhirst: @tenthirtyone Good breakdown.

contracts/factory/MagicDropCloneFactory.sol
```diff
@@ -10,21 +11,36 @@ import {MagicDropTokenImplRegistry} from "../registry/MagicDropTokenImplRegistry
...
-    =                           CONSTANTS                          =
+    =                            STORAGE                           =
     ==============================================================*/
 
+    /// @notice The registry contract
     MagicDropTokenImplRegistry private _registry;
-    bytes4 private constant INITIALIZE_SELECTOR = bytes4(keccak256("initialize(string,string,address)"));
+
+    /// @notice Gap for future upgrades
+    uint256[48] private __gap;
```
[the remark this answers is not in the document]
tomhirst: Trying to gauge context and intention here so we can move forward:

- The `MagicDropCloneFactory` implementation should be upgradable from a single proxy

- If we want to upgrade this implementation in the future, we would _not_ inherit in a `MagicDropCloneFactoryV2 is MagicDropCloneFactory` pattern, but create a new `MagicDropCloneFactoryV2`

- I can't see a use case where `MagicDropCloneFactory` would ever be inherited as a base contract (besides the our tests in `MagicDropCloneFactoryUpgradeTest.t.sol`

Therefore, if we can add any new required variables after the existing ones when upgrading implementation from `MagicDropCloneFactory` to `MagicDropCloneFactoryV2` — I'm not sure what the benefit of the `__gap` is?

contracts/factory/MagicDropCloneFactory.sol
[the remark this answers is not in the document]
tomhirst: I'm happy to leave this as is. `MagicDropCloneFactory` has no general need to deploy launchpads or self-serve contracts with the same address across multiple chains.

test/factory/MagicDropTokenImplRegistryUpgradeTest.t.sol
```diff
@@ -0,0 +1,164 @@
+// SPDX-License-Identifier: MIT
```
nothing0012, line 1: nice test

nothing0012 (approved):

PR #176, opened 2025-02-12
magicdrop cli updates

wolfy-nft:
- added withdraw function
- updated ownership transfer to be single step
- misc cleanup

### DIFF
--- a/cli/cmds/const
+++ b/cli/cmds/const
@@ -28,6 +28,8 @@ MAGIC_EDEN_DEFAULT_LIST_ID="1"
 # We use list 3 for Polygon because list 1 was already taken.
 MAGIC_EDEN_POLYGON_LIST_ID="3"
 
+DEFAULT_IMPL_ID="0"
+
 get_factory_address() {
     # abstract factory address
     if [ "$chain_id" == "2741" ]; then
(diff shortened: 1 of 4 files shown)


cli/cmds/const
```diff
@@ -45,3 +47,22 @@ get_registry_address() {
         echo "0x00000000caF1E3978e291c5Fb53FeedB957eC146"
     fi
 }
+
+# The latest MagicDrop v1.0.1 implementation ID for each supported chain.
+get_impl_id() {
```
nothing0012, line 52: where can we find all the impl ids?

cli/cmds/contract
```diff
@@ -920,3 +902,47 @@ set_cosigner_contract() {
...
+    echo ""
+
+    if [ "$balance" == "0" ]; then
+        echo "Contract has no balance to withdraw."
+        exit 1
+    fi
+
+    print_signer_with_balance $chain_id
+
+    if gum confirm "Do you want to withdraw the entire balance?"; then
```
nothing0012, line 930: oh, learned something about gum now, cool tool. do we need users to install gum first?

nothing0012 (approved):

cli/cmds/contract
wolfy-nft: already setup in our install script

PR #182, opened 2025-02-26
[Berachain] support transfer validator in cli

tomhirst:
Update CLI commands to support setting of transfer validator on 721C contracts on Berachain.

Also preemptively adds the Abstract transfer validator address.


nothing0012 (approved):
why are these LIMITBREAK_TRANSFER_VALIDATOR_V3_ needed for new chains?

tomhirst:
> why are these LIMITBREAK_TRANSFER_VALIDATOR_V3_ needed for new chains?

Because they aren't deployed to the same address.

PR #185, opened 2025-03-02
[Scripts] Deploy implementations direct

tomhirst:
We have a client request to deploy a launchpad contract with a vanity address, but we can't do this with our existing factory deployment flow:

The salt for the vanity address needs to be pre-computed with `forge create2`. However, when this salt is passed to `createContractDeterministic` it's re-hashed with `msg.sender` and `block.chainId`. Meaning, the salt will not be the same and the vanity address with not be achieved: https://github.com/me-foundation/magicdrop/blob/cd314bc3f1f62c96987faf399f0252a8d29cfd15/contracts/factory/MagicDropCloneFactory.sol#L122

These scripts directly deploy ERC721CM (non-initializable variant) directly using the pre-computed salt to achieve the vanity address via CREATE2.

### DIFF
--- a/contracts/nft/erc721m/ERC721CM.sol
+++ b/contracts/nft/erc721m/ERC721CM.sol
@@ -54,8 +54,11 @@ contract ERC721CM is
         uint256 timestampExpirySeconds,
         address mintCurrency,
         address fundReceiver,
-        uint256 mintFee
-    ) ERC721ACQueryable(collectionName, collectionSymbol) Ownable() {
+        uint256 mintFee,
+        address initialOwner
+    ) ERC721ACQueryable(collectionName, collectionSymbol) {
+        _transferOwnership(initialOwner);
+
         if (globalWalletLimit > maxMintableSupply) {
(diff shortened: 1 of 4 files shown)


tenthirtyone:
Cool with however you want to approach this. The smallest impact way I saw to accomplish this is to add a `nonce` to [this line](https://github.com/me-foundation/magicdrop/blob/main/contracts/factory/MagicDropCloneFactory.sol#L122) and take the parameter. Then you can grind CREATE2 addresses all day

tomhirst:
> Cool with however you want to approach this. The smallest impact way I saw to accomplish this is to add a `nonce` to [this line](https://github.com/me-foundation/magicdrop/blob/main/contracts/factory/MagicDropCloneFactory.sol#L122) and take the parameter. Then you can grind CREATE2 addresses all day

Cool. Interested to see how that would look. Couldn't we take the `_salt` parameter direct?

Isn't `nonce` only needed for CREATE?

How would you mine the CREATE2 addresses?

PR #187, opened 2025-03-03
Deploy MagicDropV1.0.2

tomhirst:
`MagicDropV1_0_2` contracts deployed to the same address on all chains (apart from Abstract where the same salts generate different addresses).

This PR:
- Unblocks the new version deployment pipeline
- Adds `ProxyInitCodeHelper` script for help with create2 proxy addresses
- Allows the pre-mining of cross-chain vanity addresses for both implementation and proxy contracts
- Updates the `MagicDropTokenImplRegistry`, `MagicDropFactory` and `MagicDropImpl` deployment scripts
- Adds zk-compatible variants of the contracts using Solady scripts that aren't safe to use on ZKsync-based chains like Abstract — namely, `LibClone`, Solady's `1155` token standard and  `SafeTransferLib`
- Adds various config and notation for deployment to chains that require specific details
- Bumps implementation version to `1.0.2` in dApp consumption functions

Note: Please feel free to ignore contracts in `zksync` folders when reviewing. They are largely simple duplications of the non-zkEVM variants with the Solady imports fixed to zk-compatible variants of the snippets used.


scripts-foundry/common/ProxyInitCodeHelper.s.sol
```diff
@@ -0,0 +1,18 @@
...
+import {LibClone} from "solady/src/utils/LibClone.sol";
+
+contract ProxyInitCodeHelper is Script {
+    function run() external {
+        address implementation = address(uint160(vm.envUint("IMPLEMENTATION")));
+       
+        vm.startBroadcast();
+        
+        bytes memory proxyInitCode = LibClone.initCodeERC1967(implementation);
+        console.log("Proxy init code: %s", vm.toString(proxyInitCode));
```
tenthirtyone, line 14: After it's console logged, then what? Is this verification? Is it copy/pasted somewhere? Is it dumped in a var and used with the cast command in another file?

tenthirtyone (approved):
Looks fine to me. I avoided the zsksync folders.

nothing0012 (approved):

PR #193, opened 2025-04-02
Chore/unit test migration

tenthirtyone:


test/erc721m/ERC721M.test.ts
```diff
@@ -77,700 +77,8 @@ describe('ERC721M', function () {
...
-      );
-      expect(contractBalancePost.sub(contractBalanceInitial)).to.equal(0);
-
-      const mintFeeReceiverBalancePost =
-        await ethers.provider.getBalance(MINT_FEE_RECEIVER);
-      expect(
-        mintFeeReceiverBalancePost.sub(mintFeeReceiverBalanceInitial),
-      ).to.equal(MINT_FEE);
-    });
+  describe('Minting', function () {
```
tomhirst, line 80: Are we refactoring piece by piece? Wondering why the legacy Hardhat test file still exists.

tenthirtyone: Ah, yeah I was doing it in pieces. I had some extra time and I don't know when/if I can get back to a full refactor. 

I left behind the cosign tests because I hadn't replicated every one of them yet. I just moved 1:1 as I could

PR #199, opened 2025-04-16
feat: integrate turnkey

rovilay:
- deprecate cast commands
- integrate turnkey
- use viem for contract call
- add totalTokens flag to deploy cmd
- avoid creating dir on deploy cmd


scripts/utils/getStagesData.ts
```diff
@@ -583,6 +657,8 @@ const parseAndValidateArgs = () => {
   const outputFileDir = process.argv[4];
   const tokenStandard = process.argv[5];
   const web3StorageKey = process.argv[6];
+  const isJsonOutput = process.argv[7] === 'true';
+  console.log('process.argv[7]🙄 ', process.argv[7], isJsonOutput);
```
tomhirst, line 661: @rovilay Can we remove?

cli-typescript/src/abis/index.ts
```diff
@@ -0,0 +1,386 @@
...
+    { name: 'contractAddress', type: 'address' },
+    { name: 'creator', type: 'address' },
+    { name: 'implId', type: 'uint32' },
+    { name: 'standardId', type: 'uint8' },
+    { name: 'name', type: 'string' },
+    { name: 'symbol', type: 'string' },
+  ],
+} as const;
+
+export const ERC721mAbis = {
```
tomhirst, line 81: nit: Let's pick a variable naming style for all ABIs and stick to it. IE, `TYPICAL_CONST_STYLE` or `camelCase`.

scripts/utils/getStagesData.ts
rovilay: Good catch. Fixed.

cli-typescript/src/abis/index.ts
rovilay: Fixed!

nothing0012 (approved):

PR #204, opened 2025-04-24
feat: initialize contract command

rovilay:
- add `init-contract` cmd

### DIFF
--- a/cli-typescript/src/utils/deployContract.ts
+++ b/cli-typescript/src/utils/deployContract.ts
@@ -1,7 +1,7 @@
 import path from 'path';
 import fs from 'fs';
 import { confirm } from '@inquirer/prompts';
-import { collapseAddress, executeCommand, saveDeploymentData } from './common';
+import { collapseAddress, executeCommand } from './common';
 import {
   confirmDeployment,
   confirmSetup,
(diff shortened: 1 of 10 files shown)


cli-typescript/src/utils/deployContract.ts
```diff
@@ -438,6 +399,75 @@ export const processStages = async (params: {
...
+    const parsedStagesData = stagesData.map((stage) => {
+      return [
+        BigInt(stage.price),
+        BigInt(stage.mintFee),
+        stage.walletLimit,
+        stage.merkleRoot as Hex,
+        stage.maxStageSupply,
+        BigInt(stage.startTime),
+        BigInt(stage.endTime),
+      ] as readonly [bigint, bigint, number, Hex, number, bigint, bigint];
```
tomhirst, line 442: Wonder why we have to double cast types here?

rovilay: The AbiFunction requires precise location-based typings, but the map cmd only returns an array of union types, hence the typecasting.

tomhirst (approved):

nothing0012 (approved):

PR #208, opened 2025-05-02
chore: move template files from json to typescript

rovilay:
Move template files from external json file to typescript. This will make it contract-works migration seamless.


tomhirst:
I'm worried that this would make new launchpad config harder for ops. Ops are used to creating JSON files for EVM contract deployment. I think it becomes non-obvious where to add new config files with this change.

Also, the bash CLI writes to the JSON config files on contract creation, IE, contract address and deployment date. It would be strange to write to a TS file.

Maybe I'm missing something.

rovilay:
> I'm worried that this would make new launchpad config harder for ops. Ops are used to creating JSON files for EVM contract deployment. I think it becomes non-obvious where to add new config files with this change.
> 
> Also, the bash CLI writes to the JSON config files on contract creation, IE, contract address and deployment date. It would be strange to write to a TS file.
> 
> Maybe I'm missing something.

Yeah, this doesn't change the current flow with the bash CLI i.e the new launchpad config is still created in JSON files within  the collections folder. What this PR does is to move all JSON hardcoded templates to TS files for easy integration. In terms of Contract-works, it enables easy integration and mirror the current config file management in CW.

tomhirst (approved):

nothing0012 (approved):

PR #219, opened 2025-05-30
chore: move getStagesData script

rovilay:


tomhirst:
Remind me where we're moving this from 🙏

rovilay:
> Remind me where we're moving this from 🙏

So we were using the [getStagesData script](https://github.com/rovilay/magicdrop/blob/07601e0b416a15c955767ba46195ac0da1295701/scripts/utils/getStagesData.ts#L596-L597) for the bash CLI but now I don't think it's necessary to have it separate from the typescript cli.

nothing0012 (approved):

PR #227, opened 2025-06-11
feat(magicdrop 2): freeze or thaw 721 contract

rovilay:


tomhirst (approved):
So you're enabling freeze/thaw on 721 contracts too, where before it was scoped only to 1155?

I see the setTransferable function here so looks good https://github.com/me-foundation/magicdrop/blob/46c99d79e09a67eac615c3485236839ce4f0e5db/contracts/nft/erc721m/ERC721CMInitializableV1_0_2.sol#L459

PR #234, opened 2025-07-22
fix(cli-typescript): useERC721C correct value

rovilay:


cli-typescript/src/utils/getters.ts
```diff
@@ -298,10 +298,6 @@ export const getStandardId = (tokenStandard: TOKEN_STANDARD): string => {
   }
 };
 
-export const getUseERC721C = (): boolean => {
-  return process.env.USE_ERC721C === 'true' ? true : false;
```
tomhirst, line 302: Is it possible for this to be a param we pass on deploying the contract rather than an env var?

tomhirst (approved):

tenthirtyone (approved):

PR #243, opened 2026-01-04
update MagicDrop1 to integrate with MegaETH chain

superma58:


tenthirtyone:
Looks good, but I think there may be an explorer that is still set to `TBD` in `constants.ts` 

```
[SUPPORTED_CHAINS.MEGAETH]: 'https://TBD', // MegaETH
```

tenthirtyone (approved):
approving, left comment but it isnt blocking imo

